PR #14, opened 2021-04-09
add rules for kappa and omega for gcv

ismailsenoz:
This pr adds kappa and omega update rules for the gcv node.

### DIFF
--- a/src/distributions/exp_linear_quadratic.jl
+++ b/src/distributions/exp_linear_quadratic.jl
@@ -10,16 +10,18 @@ struct ExponentialLinearQuadratic{ A <: AbstractApproximationMethod, T <: Real }
     d :: T
 end
 
-ExponentialLinearQuadratic(a::Real, b::Real, c::Real, d::Real)             = ExponentialLinearQuadratic(promote(a, b, c, d)...)
-ExponentialLinearQuadratic(a::Integer, b::Integer, c::Integer, d::Integer) = ExponentialLinearQuadratic(float(a), float(b), float(c), float(d))
+ExponentialLinearQuadratic(approximation, a::Real, b::Real, c::Real, d::Real)             = ExponentialLinearQuadratic(approximation, promote(a, b, c, d)...)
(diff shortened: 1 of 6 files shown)


albertpod (changes requested):
Change GCVMetadata to GCVMeta. Dispatch on specific types in rules where possible.

bvdmitri (changes requested):
We need to change the way we pass meta to GCV rules. Currently default meta is `nothing`.
`get_variance_approximation` function fallbacks to proper approximation in case of `nothing`. But we want to avoid to fallback every time we call this function. It computes weights and points every time we compute any message out of GCV node. 

Instead of `get_variance_approximation ` we should use a  `default_meta(::Type)` function. See an example in autoregressive node. It should return a proper default metadata object once. In all of the rules we need to change then `meta::Any` to `meta::GCVMeta`

**Edit**: my bad, it is **not** compute weights and points every time, but still we need to change it to be consistent with other pars of the framework.

ismailsenoz:
Thanks @bvdmitri for updating the meta objects. @albertpod I think it would be nice to retain a generic probability distribution instead of types for the computation of updates. VMP updates do not require a specific type of the distribution but rather requires expectation of some sufficient statistics. In case these are available requiring an additional type for dispatching will limit the capabilities. However I agree that this causes some issues with missing data for example as it tries to do math with missing. Any ideas? @bvdmitri @albertpod

bvdmitri:
@ismailsenoz @albertpod I think the PR is fine. `missing` may be a problem in the future. Though I think we can implement a nice solution. We may introduce our own type `Uninformative` with `uniformative` instance. We can define whatever we want for this `Uninformative` type, `mean()`, `var()`, etc... Basically all of this methods should just throw an error with descriptive error message. 

Doing the same with `missing` type would be a type-piracy in Julia, which is considered to be a bad practice.
https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-type-piracy

Besides that, nicely done with this PR, I'm going to merge it, thanks @ismailsenoz !

bvdmitri (approved):

PR #16, opened 2021-04-17
Develop gcv params

ismailsenoz:
This PR implements mean-field update rules and mean-field average energy function for the gcv node.


bvdmitri (approved):
Thanks @ismailsenoz !

Issue #15, opened 2021-04-15
Generalise Gaussian rules to support all Gaussian parametrisations


bvdmitri:
Currently nodes from Gaussian distributions family accept messages of the same parametrisation only. 
E.g.
```
@rule NormalMeanPrecision(:μ, Marginalisation) (m_out::NormalMeanPrecision, q_τ::Any) = NormalMeanPrecision(mean(m_out), cholinv( cov(m_out) + cholinv(mean(q_τ)) ))
```

We need to generalise it with `UnivariateNormalDistributionsFamily`

PR #17, opened 2021-05-24
Develop rules gmv

bvdmitri:
This PR implements several important things:

1. Generic functional message in the form of `ContinuousUnivariateLogPdf` distribution. This distribution supports various domains and provides an interface for further message approximation (has not been implemented yet)
2. Testing utilities for testing rules and their results
3. As an example of introduced changes almost all rules have been reimplemented and tested for `NormalMeanVariance` node


bvdmitri:
Related to #15

PR #18, opened 2021-05-27
extend update rules for mvnormal

bartvanerp:
This PR adds update rules for the multivariate normal node. It optimizes some rules to potentially reduce the number of matrix inversions. Furthermore it includes tests for the added rules.

### DIFF
--- a/src/rule.jl
+++ b/src/rule.jl
@@ -266,6 +266,11 @@ macro test_rules(options, on, test_sequence)
     @capture(options, [ option_entries__ ]) || error("Invalid options specification. Options should be in the form on an array.")
     
     with_float_conversions = false
+
+    float64_atol  = 1e-12
+    float32_atol  = 1e-6
+    bigfloat_atol = 1e-12
+
     
     foreach(option_entries) do option_entry 
         @capture(option_entry, (key_ = value_)) || error("Invalid option entry specification: $(option_entry). Option entry should be in the form of a 'key = value' pair.")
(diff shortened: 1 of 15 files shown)


bartvanerp:
Some of the original functions have also been updated for speed. In some cases a `MvNormalMeanPrecision` was returned, which required another matrix inversion to convert the calculated covariance matrix to a precision matrix. These rules now return `MvNormalMeanCovariance` to prevent this matrix inversion.

One issue arose whilst implementing the tests. For `mv_normal_mean_precision/test_precision.jl` the options `with_float_conversions` had to be set to `false`, because the Wishart distribution was converted to Float32. I am not sure whether this is desired behaviour.

bvdmitri:
@bartvanerp thank you very much for this PR. I will review it on the next week.

Regarding your question:

> because the Wishart distribution was converted to Float32. I am not sure whether this is desired behaviour.

It is desired behaviour. It is better to test rules with `with_float_conversions = true` unless it is technically impossible. The idea behind this option is that testing function tries to input messages to a different float precisions (for all possible combinations, currently to Float32 and BigFloat) and checks the output type. The rules are the following:

- If all inputs are Float32 the output is Float32
- if some inputs are Float32 the output is Float64 (because Float64 is used by default for remaining distributions, type promotion to Float64)
- if some inputs are BigFloat the output is BigFloat (type promotion to BigFloat)
- if all inputs are BigFloat the output is BigFloat 

This options ensures correct type promotion and whether rules are generic enough or not. In principle it also allows to check (in most of the cases) that rule is differentiable and promotes types correctly for `<: Real` values. E.g. `ForwardDiff` uses `Dual` numbers which are `<: Real` and for correct gradients the output has to be `Dual`. This testing function is undocumented yet so its fine for now, you don't have to fix it. I will check later on why it may cause some problem with `Wishart` distribution.

PR #24, opened 2021-07-09
Probit node ported to ReactiveMP

bartvanerp:
This pull request ports the Probit node from ForneyLab.jl to ReactiveMP.jl

This node is based on the paper:
M. Cox and B. De Vries, "Robust Expectation Propagation in Factor Graphs Involving Both Continuous and Binary Variables," 2018 26th European Signal Processing Conference (EUSIPCO), 2018, pp. 2583-2587, doi: 10.23919/EUSIPCO.2018.8553490.
available [here](https://biaslab.github.io/pdf/eusipco2018/cox_robust_expectation_propagation.pdf). 

Importantly, the derivations from the paper and the implementation differ slightly as the second (unnormalized!) moment of g (from the paper) has one of its signs inverted. This has been fixed in the implementation.

Additionally, the calculation of the average energy of the Probit node has been altered with respect to its implementation in ForneyLab. In the case that we have a standard Normal distribution on the continuous input of the node, the average energy of the node should return 1. See comment on [StackExchange](https://math.stackexchange.com/questions/2342234/does-the-expectation-of-the-log-of-a-gaussian-cumulative-distribution-function-w).

Furthermore this pull request includes tests for the Probit node and rules. Currently, the rules experience some type conversion issues, causing some of them to fail. I was not able to fix this myself.


albertpod:
Great work, @bartvanerp! Thanks for the tests! I think this PR is ready to be merged.

bvdmitri:
@bartvanerp thank you for this PR! It looks great, I fixed tests, all are green! Would it be ok for you if I wait with merging? Technically `Probit` works under moment matching constraint and not under `Marginalisation` constraint. Current implementation uses `Marginalisation` constraint however. I will work on `MomentMatching` constraint and EP algorithm in general this week and then I will merge!

src/nodes/probit.jl
```diff
@@ -0,0 +1,45 @@
...
+ProbitMeta(; p = 32) = ProbitMeta(p)
+
+@node Probit Stochastic [ out, in ]
+
+default_meta(::Type{ Probit }) = ProbitMeta(32)
+
+default_functional_dependencies_pipeline(::Type{ <: Probit }) = RequireInboundFunctionalDependencies((2, ), (vague(NormalMeanPrecision), ))
+
+default_interface_local_constraint(::Type{ <: Probit }, edge::Val{ :in })  = MomentMatching()
+default_interface_local_constraint(::Type{ <: Probit }, edge::Val{ :out }) = Marginalisation()
```
bvdmitri, line 22: @bartvanerp Review this and I will merge

bartvanerp: The changes look good. I added a final update rule to deal with missing data.

Two problems occurred for the tests:
1) It is not clear how to test for Missing inputs
2) The moment matching constraint is not implemented for the test macro I believe

bvdmitri: @bartvanerp To address your questions

1) Lets not implement `missing` rules for now, we don't have them for any node yet (only defined in notebooks as an example), I'm still not sure how it should properly work, for now user can define `missing` rules in their notebooks to make it work, but mathematical meaning of such rules has to be discussed yet.
2) yeap, we just forgot to `export MomentMatching`, that is fixed

bvdmitri:
Great work, @bartvanerp!

PR #25, opened 2021-07-12
add Aqua tests, get them to pass

cscherrer:
[Aqua.jl](https://github.com/JuliaTesting/Aqua.jl) is a package that makes it easy to do some consistency checks.

This PR adds Aqua as a test dependency, and makes some small changes to make Aqua happy.

Note that ambiguity tests are disabled. Currently there are 148 ambiguities, for example
```julia
Ambiguity #148
make_node(fform::Function, autovar::ReactiveMP.AutoVar, args::ReactiveMP.ConstVariable...; kwargs...) in ReactiveMP at /home/chad/git/ReactiveMP.jl/src/node.jl:683
make_node(fform::typeof(+), autovar::ReactiveMP.AutoVar, args::ReactiveMP.AbstractVariable...; kwargs...) in ReactiveMP at /home/chad/git/ReactiveMP.jl/src/node.jl:845
```


bvdmitri:
Thanks @cscherrer!

PR #30, opened 2021-07-31
Extension rule addition

A-Vzer:
Added the addition rules for several distributions. 

If precision is the first argument (ignoring 'nothing') then it returns the precision parameterized distribution.

Test scripts work. If this part is correct I will move on to the rules of the normal distributions.

Should the test scripts be updated as well ?


bartvanerp:
Hi Amaan,

Thanks for the first PR. We really appreciate the effort. 

We have some comments regarding the PR (it is quite extensive, but it is meant as a complete overview for following PRs):
- Indeed we would also like you to implement the test scripts. Not only will they make sure that the package will become less error-prone, it will also check whether the new update rules are implemented correctly. For each (typed) rule we would like to see about 3 tests.
- Be careful with the return type when there is a `NormalDistributionsFamily` as input, such as in
https://github.com/biaslab/ReactiveMP.jl/blob/c801283940a74ac56ed2802b689b74764a8cb289/src/rules/addition/in1.jl#L3
The `NormalDistributionsFamily` is the union of the `UnivariateNormalDistributionsFamily` and `MultivariateNormalDistributionsFamily` types. As these have different dimensionalities (univariate vs multivariate), we cannot immediately infer what the dimension of the output should be and therefore we would not know directly whether to return `NormalMeanVariance()` or `MvNormalMeanCovariance()`. As a fix we would recommend to only specify ruls for  `UnivariateNormalDistributionsFamily` and  `MultivariateNormalDistributionsFamily` (or subtypes) and not for `NormalDistributionsFamily`.
- Make sure to use the `cov()` function for multivariate normal distributions instead of the `var()` function, as `var()` will only return the diagonal of the covariance matrix, e.g.
https://github.com/biaslab/ReactiveMP.jl/blob/c801283940a74ac56ed2802b689b74764a8cb289/src/rules/addition/in1.jl#L16
As a habit we would recommend to use `cov()` exclusively for multivariate distributions and `var()` exclusively for univariate distributions.
- As a habit try to always use `mean_cov()` (or similar) when you need multiple statistics of the normal distribution, e.g. 
https://github.com/biaslab/ReactiveMP.jl/blob/c801283940a74ac56ed2802b689b74764a8cb289/src/rules/addition/in1.jl#L6
would then become
```julia
@rule typeof(+)(:in1, Marginalisation) (m_out::PointMass, m_in2::MultivariateNormalDistributionsFamily) = begin

  # extract parameters
  min2, vin2 = mean_cov(m_in2)
  
  # return distribution
  return MvNormalMeanCovariance(mean(m_out) - min2, vin2)

end

```
- Keep in mind that in the `MvNormalPrecision` type we have access to the precision matrix (without requiring inverses) and in the `MvNormalMeanCovariance` type we have access to the covariance matrix. Therefore some addition rules can be further optimized such as 
https://github.com/biaslab/ReactiveMP.jl/blob/c801283940a74ac56ed2802b689b74764a8cb289/src/rules/addition/in1.jl#L7
to
```julia
@rule typeof(+)(:in1, Marginalisation) (m_out::PointMass, m_in2::MvNormalMeanPrecision) = begin

  # extract parameters
  min2, win2 = mean_precision(m_in2)
  
  # return distribution
  return MvNormalMeanPrecision(mean(m_out) - min2, win2)

end

```
- The `cov()` and `precision()` functions just extract the corresponding matrices from the multivariate distributions. Conversion of these matrices should be performed using `cholinv()` (or by 1/... for scalar values). Although in this node we do not need it as we can change the return type. If we then later on need its inverse, then ReactiveMP will automatically handle this, but in this way we at least prevent inversion within this rule e.g. 
https://github.com/biaslab/ReactiveMP.jl/blob/c801283940a74ac56ed2802b689b74764a8cb289/src/rules/addition/in1.jl#L54
```julia
@rule typeof(+)(:in1, Marginalisation) (m_out::MvNormalMeanPrecision, m_in2::MvNormalMeanPrecision) = begin

  # extract parameters
  mout, vout = mean_cov(m_out)
  min2, vin2 = mean_cov(m_in2)
  
  # return distribution
  return MvNormalMeanCovariance(mout-min2, vout + vin2)

end
```

bartvanerp (approved):

bvdmitri (approved):
Well done

PR #33, opened 2021-08-06
feat(): getmarginal support for joint marginals around some node

bvdmitri:
This PR implements `getmarginal` function for factor node that supports indexing marginal by their names.


bartvanerp (approved):
Works like a charm. I also added the getmarginals function for arrays of factornodes.

PR #35, opened 2021-08-11
Integration of BIFM node in ReactiveMP

bartvanerp:
This pull request contains:
1. A `BIFM` node with update rules.
2. A `BIFM_helper` node with update rules.
3. A demo notebook, with benchmarking comparison.
4. The `NormalWeightedMeanPrecision` node without update rules (to be used just as a prior).
5. A so-called `MarginalDistribution` wrapper around a regular distribution, required for performing BIFM smoothing through message passing. It allows to propagate marginals directly over the graph as messages. The `prod` function used for creating marginals simply only keeps the message of type `MarginalDistribution`.

Changes with respect to the last version (June 2021):
- `functional_dependencies` of `BIFM_helper` node updated to work with current master branch.
- Update rule towards the input has been changed according to Wadehn 2019.
- Code refactored/slightly optimized.

This notebook is based on:
F. Wadehn, “State Space Methods with Applications in Biomedical Signal Processing,” ETH Zurich, 2019. Accessed: Jun. 16, 2021. [Online]. Available: https://www.research-collection.ethz.ch/handle/20.500.11850/344762
H. Loeliger, L. Bruderer, H. Malmberg, F. Wadehn, and N. Zalmai, “On sparsity by NUV-EM, Gaussian message passing, and Kalman smoothing,” in 2016 Information Theory and Applications Workshop (ITA), Jan. 2016, pp. 1–10. doi: 10.1109/ITA.2016.7888168.

Credits to Martin the Quincey for implementation and experiments.


Martindeq98 (approved):
I only checked the files regarding BIFM, I explicitly checked the files where I have also checked the box "viewed".

Some nice improvements have been made to the BIFM node, good job! It is nice to see that e.g. the functional dependency modifications are now handled inside the node rather than inside the Jupyter Notebook. The mathematical model at the start of notebook "RTS vs BIFM Smoothing.ipynb" is not correct: The X_k is not defined in the second equation, and the Z_k is added incorrectly in the second equation. Also, the interchanging of the variable for the latent state could be confusing for some; X and Z are sometimes used interchangeably throughout the code and the text.

The inference of the latent states seems good, in some dimensions it is slightly different than RTS but I think this is not a problem. However, it looks like the input marginals for the two smoothers are not completely agreeing with each other (see RTS vs BIFM Smoothing.ipynb, input estimation plot). I had not fully checked this at the end of the internship due to time constraints, so I am not 100% sure it is correct.

So overall, everything seems very good! I just express some of my uncertainties here.

bartvanerp:
Hi @Martindeq98 ,

Thanks a lot for your time and the review! 

Indeed I accidentally merged the variable names from two distinct notations, thanks for the catch. The latest commit changes these variable names throughout the entire demo, but does not affect the inference results.

With respect to the input estimation I agree that there is a discrepancy there, which is both visible in the state estimate and input estimation plots. These discrepancies occur around the same time indices, indicating that one causes another (and vice versa). I double-checked the updates and these seem fine. After running the experiments a couple of times, I also noted that the RTS smoother sometimes fails to infer the latent state correctly as it deviates from the actual hidden state. This also resulted in differences in input estimation. In contract, the BIFM smoother did seem to infer the latent state correctly. As the latent state in the experiments is 25 and we encountered numerical stability issues before, I assume that this is what is causing the discrepancy. However, it is important to note that the RTS smoother is the cause of this.

Besides this I see that I forgot the marginal rules for free energy calculation and tests, these will follow shortly.

bvdmitri (approved):

PR #38, opened 2021-08-17
fix incoming message types and mean-cov methods for function.jl

ismailsenoz:


albertpod:
is this PR still relevant @bvdmitri?

bvdmitri:
@albertpod No, I don't think so. Probably can be closed. We fixed most of the issues in another PR. Regarding `mean`, `cov` functions for `ContinuousUnivariateLogPdf ` I would say its better to use form constrains, constraint it to be Gaussian with suitable approximation method and only then call `mean`.

PR #44, opened 2021-09-18
dot product rules

A-Vzer:
few points:

- should the dot product rule not be under the "deterministic" heading in the ` ReaciveMP.jl` file?
- I imported "dot" from LinearAlgebra to make the node work
- I put the constraint in the test_rule identical as is done in the rule definition but the following error comes:
`Error in macro. Functional form specification should in the form of 'fformtype_(on_, vconstraint_)'`


bartvanerp:
Thanks for starting this PR. With respect to your questions/remarks:
- Indeed, the dot node should be included under the deterministic nodes section.
- I think it is good that you explicitly import `dot` from `LinearAlgebra`. This makes sure that the node and the dot product have the same command and that there won't be a clash when importing/using the `LinearAlgebra` package. However, it is kind of strange to me why the `dot` node worked before. Do you know this @bvdmitri? Perhaps we include/import it in some other file?
- The error that you report has to do with the `symmetric` argument in the rule, which has not been implemented in the test rules. For the test rules it is best to remove the `symmetric` argument in the test rule (not the actual rule) and to test both cases separately (just to make sure that the `symmetric` argument also works). In your example this would mean testing for both the combinations `PointMass - Normal` and `Normal - Pointmass`. If you remove this argument in the test that is currently implemented, then it will still throw an error, because it currently does not know about the dotproduct of a scalar and a normal. Let's also implement this rule for the sake of generality.

A-Vzer:
Thanks for the feedback! Ive change some rules to more specific cases so that I could avoid using symmetrical. Also I added rules that do not include `meta` as an argument. let me know if this is good or not

src/rules/dot_product/in1.jl
```diff
@@ -4,5 +4,40 @@ export rule
...
+    x  = mean(m_in2)
+    xi = x * weightedmean(m_out)
+    w  = x * precision(m_out) * x'
+    return NormalWeightedMeanPrecision(xi, w)
+end
+
+@rule typeof(dot)(:in1, Marginalisation) (m_out::PointMass{ <: AbstractVector }, m_in2::UnivariateNormalDistributionsFamily, meta::AbstractCorrection) = begin
+    x  = mean(m_out)
+    xi = x * weightedmean(m_in2)
+    w  = correction!(meta, x * precision(m_in2) * x')
```
bartvanerp, line 27: Please combine `weighedmean(m_in2)` and `precision(m_in2)` into `weightedmean_precision(m_in2)` for possibly improved performance, also for the other rules.

src/rules/dot_product/in1.jl
```diff
@@ -4,5 +4,40 @@ export rule
...
+
+@rule typeof(dot)(:in1, Marginalisation) (m_out::UnivariateNormalDistributionsFamily, m_in2::PointMass, meta::AbstractCorrection) = begin
+    x  = mean(m_in2)
+    xi = x * weightedmean(m_out)
+    w  = correction!(meta, x * precision(m_out) * x')
+    return NormalWeightedMeanPrecision(xi, w)
+end
+
+@rule typeof(dot)(:in1, Marginalisation) (m_out::UnivariateNormalDistributionsFamily, m_in2::PointMass) = begin
+    x  = mean(m_in2)
```
bartvanerp, line 18: Good that you added additional rules, however, in this case it is not desired. The `meta` argument specifies how the precision should be corrected. The `dot` node always has this object (see `dot_product.jl`) which defaults to `TinyCorrection`. So we are safe to assume that it is always available for our rules. Without this correction we might see problems such as numerical stability when performing the calculations. Therefore please remove the rules without the `meta` argument.

test/rules/dot_product/test_in2.jl
```diff
@@ -0,0 +1,32 @@
...
+
+using Test
+using ReactiveMP
+using Random
+import ReactiveMP: @test_rules
+import LinearAlgebra: dot
+
+@testset "rules:typeof(dot):in2" begin
+
+     @testset "Belief Propagation: (m_out::UnivariateNormalDistributionsFamily, m_in1::PointMass)" begin
```
bartvanerp, line 11: I am not sure whether the `meta` argument is automatically passed to the test function. @bvdmitri could you verify this?

bvdmitri (approved):

PR #45, opened 2021-09-18
Exponential distirbution/node rules

A-Vzer:
I tried implementing the exponential rule however it does not recognize the rule i have created and throws
`RuleMethodError: no method matching rule for the given arguments` even though the rule is clearly (to me) defined.


bartvanerp:
Thanks for starting this PR. The error was caused because the `out.jl` file for the `Exponential` node was not included under `prototypes.jl`. Hence, the package did not load the rule and therefore for the test it was unknown. I quickly reordered the rules to alphabetical order in the `prototypes.jl` file for clarity. Please also don't forget to include the test in the `runtests.jl` file (and possibly reorder these too).

src/rules/exponential/out.jl
```diff
@@ -0,0 +1,6 @@
+export rule
+using Distributions
+
+@rule Exponential(:out, Marginalisation) (m_in1::PointMass, ) = begin
+    return PointMass(exp(mean(m_in1)))
```
bvdmitri, line 5: I don't think this is correct. It should be
```
return Exponential(mean(m_in1))
```

But I can see indeed that in `ForneyLab` its just `exp`. What do you think @bartvanerp? It seems to me that in ForneyLab `Exponential` node is not a distribution, but a deterministic `exp` function. It has the same name but the meaning is not quite the same, right?

EDIT: My assumption is correct, in ForneyLab its just an `exp` mapping, see [here](https://github.com/biaslab/ForneyLab.jl/blob/master/src/factor_nodes/exponential.jl). We will think about it with @bartvanerp. You can skip this one for now @A-Vzer 

Also better to avoid uninformative names in actual distributions, for [Exponential distribution](https://en.wikipedia.org/wiki/Exponential_distribution) instead of `in1` its better to use `\lambda<TAB>`

bartvanerp:
@bvdmitri Can we close this PR? I think the exponential mapping will be resolved with the `Delta` node. If someone needs an `Exponential` distribution, then we will create an issue later.

bvdmitri:
@bartvanerp 👍 yes, I think we can close this

PR #46, opened 2021-09-22
fix(): Fix for GaussianMixture Node average energy function

bvdmitri:
This PR fixes average energy score function signature for Gaussian Mixture node so it can work with different gaussian parametrisation.


albertpod (approved):
Well-spotted @bvdmitri!

PR #47, opened 2021-09-23
feat(): extra Gaussian nodes family rules and performance improvements

bvdmitri:
This PR fixes `GaussianMeanPrecision` node rules to handle all Gaussian parametrisation cases.

Thanks to @ismailsenoz

### DIFF
--- a/src/rules/normal_mean_precision/mean.jl
+++ b/src/rules/normal_mean_precision/mean.jl
@@ -12,4 +12,7 @@ end
 # --------------------------------- #
 @rule NormalMeanPrecision(:μ, Marginalisation) (q_out::Any, q_τ::Any) = NormalMeanPrecision(mean(q_out), mean(q_τ))
 
-@rule NormalMeanPrecision(:μ, Marginalisation) (m_out::UnivariateNormalDistributionsFamily, q_τ::Any) = NormalMeanPrecision(mean(m_out), cholinv( cov(m_out) + cholinv(mean(q_τ)) ))
+@rule NormalMeanPrecision(:μ, Marginalisation) (m_out::UnivariateNormalDistributionsFamily, q_τ::Any) = begin 
+    m_out_mean, m_out_cov = mean_cov(m_out)
+    return NormalMeanPrecision(m_out_mean, cholinv(m_out_cov + cholinv(mean(q_τ)) ))
+end
--- a/src/rules/normal_mean_precision/mean.jl
+++ b/src/rules/normal_mean_precision/mean.jl
@@ -5,7 +5,7 @@
 
 @rule NormalMeanPrecision(:μ, Marginalisation) (m_out::UnivariateNormalDistributionsFamily, m_τ::PointMass) = begin 
     m_out_mean, m_out_cov = mean_cov(m_out)
-    return NormalMeanPrecision(m_out_mean, inv(m_out_cov + inv(mean(m_τ))))
+    return NormalMeanVariance(m_out_mean, m_out_cov + inv(mean(m_τ)))
 end
 
 # Variational                       # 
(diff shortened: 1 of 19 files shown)


albertpod (approved):

src/rules/normal_mean_precision/mean.jl
```diff
@@ -12,4 +12,7 @@ end
 # --------------------------------- #
 @rule NormalMeanPrecision(:μ, Marginalisation) (q_out::Any, q_τ::Any) = NormalMeanPrecision(mean(q_out), mean(q_τ))
 
-@rule NormalMeanPrecision(:μ, Marginalisation) (m_out::NormalMeanPrecision, q_τ::Any) = NormalMeanPrecision(mean(m_out), cholinv( cov(m_out) + cholinv(mean(q_τ)) ))
\ No newline at end of file
+@rule NormalMeanPrecision(:μ, Marginalisation) (m_out::UnivariateNormalDistributionsFamily, q_τ::Any) = begin 
+    m_out_mean, m_out_cov = mean_cov(m_out)
+    return NormalMeanPrecision(m_out_mean, cholinv(m_out_cov + cholinv(mean(q_τ)) ))
```
bartvanerp, line 17: Why do we return a precision matrix here? A covariance matrix would make more sense as this requires one less inversion. If an inversion is required later on, then this can still be performed.

src/rules/normal_mean_precision/out.jl
```diff
@@ -12,4 +12,7 @@ end
 # --------------------------------- #
 @rule NormalMeanPrecision(:out, Marginalisation) (q_μ::Any, q_τ::Any) = NormalMeanPrecision(mean(q_μ), mean(q_τ))
 
-@rule NormalMeanPrecision(:out, Marginalisation) (m_μ::NormalMeanPrecision, q_τ::Any) = NormalMeanPrecision(mean(m_μ), cholinv( cov(m_μ) + cholinv(mean(q_τ)) ))
\ No newline at end of file
+@rule NormalMeanPrecision(:out, Marginalisation) (m_μ::UnivariateNormalDistributionsFamily, q_τ::Any) = begin 
+    m_μ_mean, m_μ_cov = mean_cov(m_μ)
+    return NormalMeanPrecision(m_μ_mean, cholinv(m_μ_cov + cholinv(mean(q_τ))))
```
bartvanerp, line 17: See previous comment.

bartvanerp (approved):
Good catch!

ismailsenoz (approved):

src/rules/normal_mean_precision/mean.jl
bvdmitri: That is true. In the beginning of this package idea was that a node should always return the same parametrisation as a message. However, I don't really think it is that important and we can return `NormalMeanVariance` here indeed. What do you think @albertpod? I think we have the same for multivariate case as well (extra inversion just to return a precision matrix).

albertpod: It's hard to tell which parametrization is preferable in advance.
The rule of thumb is to return a message that has fewer inversions, so in this example, `NormalMeanVariance` is a better option.
On the other hand, we know that the multiplication of Gaussians can be optimized by using their "natural parameterization". So if a Gaussian rule can be represented by weighted-mean-precision (wmp) with fewer inversions, then better to do so. `ReactiveMP.jl` casts the factors of products of two Gaussians to wmp.

bvdmitri: I think we should not assume in advance that we are going to do a multiplication of Gaussians. Even if we do, it doesn't save us even a single inversion:

- Strategy 1: Single inverse in rule - no inversions in prod
- Strategy 2: No inversions in rule - Single inverse in prod

In that sense, I thing `NormalMeanVariance` might be better if we don't do `prod` after and also doesn't affect performance if we do `prod` right after.

PR #48, opened 2021-09-29
Implement normalizing flows in ReactiveMP

bartvanerp:
This PR allows us to implement normalizing flows in ReactiveMP.

It contains the following features:
- Easy to use flow model specification.
- Additive coupling layers that can operate for any input dimension larger than 1.
- Planar and radial flow functions for describing non-linearities.
- Extensibility for custom layers with and coupling flows.
- Input layers and permutation layers for a more versatile model specification.
- Inference through both linearization and the unscented transform.
- Tutorial notebook for using these flows.
- Tests for all above features.

Changes that require further attention:
- Certain functions from the testing environment have been moved from `runtests.jl` to `test_helpers.jl`. This allows us to include the `test_helpers.jl` file in the flow node, such that the tests corresponding to the flow node can be included from `test_flow.jl` using the `addtests()` function. This prevents the `runtests.jl` file from becoming cluttered when more layers/flows are developed.

Open improvements include (will create issues for these later if that is okay):
- Generalize the partition dimension argument to work with partitions that have a dimensionality larger than 1.
- Allow for specifying custom mapping in layers through the model specification.
- Add custom initializer functions that adequately deal with domain constraints of parameters. This should enable us to sample plausible parameters from the uncompiled model.


src/algebra/permutation_matrix.jl
```diff
@@ -0,0 +1,242 @@
...
+        ind[1] = tmp
+    end
+    return PermutationMatrix(ind)
+end
+
+
+# extensions of base functionality
+Base.eltype(::PermutationMatrix{T}) where { T } = T
+Base.size(mat::PermutationMatrix)               = (length(mat), length(mat)) 
+Base.length(mat::PermutationMatrix)             = length(mat.ind)
```
bvdmitri, line 46: `Base.length(matrix::AbstractMatrix)` usually returns `prod(size(matrix))`. So for `3x3` matrix `length` should be equal to `9`.

src/algebra/permutation_matrix.jl
```diff
@@ -0,0 +1,242 @@
...
+    A_{ij} = \{0, 1\}\\
+    ∑_{i} A_{ij} = 1\\
+    ∑_{j} A_{ij} = 1
+```
+An example is the 3-dimensional permutation matrix
+```math
+A = \begin{bmatrix} 1 & 0 & 0 \\ 0 & 0 & 1 \\ 1 & 0 & 0 \end{bmatrix}
+```
+"""
+struct PermutationMatrix{ T <: Int } <: AbstractMatrix{T}
```
bvdmitri, line 21: `T <: Int` restricts `T` to be only `Int` because `Int` is a concrete type. Usually `<:` operator is used with abstract types such as `Integer` or `AbstractVector` or `Number`. `T <: Int` does not make a lot of sense in this context as it is equivalent to just `Int`

src/algebra/permutation_matrix.jl
```diff
@@ -0,0 +1,242 @@
...
+        return zero(eltype(mat))
+    end
+end
+
+LinearAlgebra.inv(mat::PermutationMatrix)       = adjoint(mat)
+
+
+# get functions
+getind(mat::PermutationMatrix)      = mat.ind
+getindices(mat::PermutationMatrix)  = mat.ind
```
bvdmitri, line 61: Do we need two identical functions here? Maybe remove one of them

src/algebra/permutation_matrix.jl
bvdmitri: But be careful as I see you use `length` later on with a different assumption. Maybe its fine to keep it as it is now.

src/nodes/flow/coupling_flows/planar_flow.jl
```diff
@@ -0,0 +1,258 @@
...
+    # return result
+    return result
+
+end
+forward(f::PlanarFlow{T1,T2}, input::T1) where { T1, T2 <: Real } = _forward(f, input)
+Broadcast.broadcasted(::typeof(forward), f::PlanarFlow{T1,T2}, input::Array{T1,1}) where { T1, T2 <: Real } = broadcast(_forward, Ref(f), input)
+
+
+# forward pass through the PlanarFlow function (univariate input)
+function _forward(f::PlanarFlow{T1,T2}, input::T3) where { T1 <: Real, T2 <: Real, T3 <: Real }
```
bvdmitri, line 120: Just in case, if you do not enforce any additional constraints on type parameters and do not use it in function body you can write:
```
function _forward(f::PlanarFlow{ <: Real, <:Real }, input::Real)

end
```

Saves you a bit of time on typing all of this. But I'm totally fine with the current version as well.

src/nodes/flow/coupling_flows/planar_flow.jl
```diff
@@ -0,0 +1,258 @@
...
+    result = copy(u)
+    result *= tanh(dot(w, input) + b)
+    result += input
+
+    # return result
+    return result
+
+end
+forward(f::PlanarFlow{T1,T2}, input::T3) where { T1 <: Real, T2 <: Real, T3 <: Real } = _forward(f, input)
+Broadcast.broadcasted(::typeof(forward), f::PlanarFlow{T1,T2}, input::Array{T3,1}) where { T1 <: Real, T2 <: Real, T3 <: Real } = broadcast(_forward, Ref(f), input)
```
bvdmitri, line 135: I would change concrete types everywhere, e.g. from `Array{T3, 1}` to `AbstractArray{T3, 1}` (or even `AbstractVector{T3}` but its minor, `AbstractArray` is also fine)

src/nodes/flow/coupling_flows/planar_flow.jl
```diff
@@ -0,0 +1,258 @@
...
+
+end
+
+# extra utility function (multivariate)
+inv_jacobian(f::PlanarFlow{T1,T2}, input::T1) where { T1, T2 <: Real }       = inv(jacobian(f, input))
+absdet_jacobian(f::PlanarFlow{T1,T2}, input::T1) where { T1, T2 <: Real }    = abs(det_jacobian(f, input))
+logabsdet_jacobian(f::PlanarFlow{T1,T2}, input::T1) where { T1, T2 <: Real}  = logabsdet(jacobian(f, input))
+
+# extra utility functions (univariate)
+inv_jacobian(f::PlanarFlow{T,T}, input::T) where { T <: Real }               = 1.0 / jacobian(f, input)
```
bvdmitri, line 250: Be careful here. So I noticed that in many places you use for example `{T1, T2} where { T1 <: Real, T2 <: Real }`. This form accepts `{Float64, Float32}` for example, because both type arguments have different names. However when you type `{T, T} where { T <: Real }` it assumes that both of the **are the same**. So some of your functions accepts different types from `Real` subtype.

For example I do not see any conceptual differences in these
```
inv_jacobian(f::PlanarFlow{T1,T2}, input::T1) where { T1, T2 <: Real }       = inv(jacobian(f, input))

inv_jacobian(f::PlanarFlow{T,T}, input::T) where { T <: Real }               = 1.0 / jacobian(f, input)
```

First version is a more generic version of a second one, but for some reason they have different implementations. I think there should not be two different implementations for `{Float32, Float64}` and `{Float64, Float64}`. Maybe it is a bug and there should be something else? I would just delete versions with `{T, T}` as they are doing exactly the same but just more restrict on their input arguments.

src/nodes/flow/coupling_flows/planar_flow.jl
```diff
@@ -0,0 +1,258 @@
...
+function _jacobian(f::PlanarFlow{T1,T2}, input::T3) where { T1 <: Real, T2 <: Real, T3 }
+    # function when the input is an array with 1 element
+    @assert length(input) == 1 "Something is wrong with the dimensionality of the input to the PlanarFlow flow."
+    return jacobian(f, input[1])
+end
+jacobian(f::PlanarFlow{T1,T2}, input::T3) where { T1 <: Real, T2 <: Real, T3 } = _jacobian(f, input)
+Broadcast.broadcasted(::typeof(jacobian), f::PlanarFlow{T1,T2}, input::Array{T3,1}) where { T1 <: Real, T2 <: Real, T3 } = broadcast(_jacobian, Ref(f), input)
+
+# inplace jacobian of the PlanarFlow function (multivariate input)
+function jacobian!(output::Array{T2,2}, f::PlanarFlow{T1,T2}, input::T1) where { T1, T2 <: Real}
```
bvdmitri, line 206: Same as previous comment, better to replace all concrete types from Base to their abstract equivalents, such as `AbstractArray{T2, 2}` or `AbstractMatrix{T2}` (it is an alias for `AbstractArray{T2 , 2}`).

src/algebra/permutation_matrix.jl
cscherrer: I'd be very concerned about leaving as is. `length` is used in lots of places in Base, and having a nonstandard definition will very likely lead to weird behavior and subtle bugs down the road. If other code comes to depend on the nonstandard implementation, it can be very hard to back out later.

src/nodes/flow/coupling_flows/planar_flow.jl
bvdmitri: Ah, I see now, that `T` is not constrained to be `Real`. They are different indeed, my bad.

src/algebra/flatten_tuple.jl
```diff
@@ -0,0 +1,6 @@
+# type stable flatten function for tuples
+flatten_tuple() = ()
+flatten_tuple(a) = (a,)
+flatten_tuple(a::Tuple, b...) = tuple(flatten_tuple(a...)..., flatten_tuple(b...)...)
+flatten_tuple(a, b...) = tuple(a, flatten_tuple(b...)...)
+flatten_tuple(x::Tuple) = tuple(flatten_tuple(x...)...)
```
cscherrer, line 6: This can be implemented more simply as
```suggestion
flatten(x, y...) = (flatten(x)..., flatten(y...)...)
flatten(x::Tuple) = flatten(x...)
flatten(x) = (x,)
```

src/algebra/helpers.jl
```diff
@@ -5,4 +5,17 @@ import LinearAlgebra: Diagonal
 xT_A_x(x::AbstractVector, A::AbstractMatrix) = dot(x, A * x)
 
 diageye(::Type{T}, n::Int) where { T <: Real } = Matrix{T}(I, n, n)
-diageye(n::Int)                                = diageye(Float64, n)
\ No newline at end of file
+diageye(n::Int)                                = diageye(Float64, n)
+
+function softmax(x::Array{Float64,1})
```
bvdmitri, line 10: We can add `LogExpFunctions.jl` dependency with `softmax` and `sigmoid` functions included. Not sure about `dtanh`. Didn't find any equivalent in other libraries

src/algebra/permutation_matrix.jl
```diff
@@ -0,0 +1,242 @@
+export PermutationMatrix, PT_X_P, PT_X_P!, getind ,getindices
```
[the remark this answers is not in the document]
bvdmitri: Please check everywhere and export only needed items. Other functions/structures might be package local.

src/nodes/flow/flow_models/flow_model.jl
```diff
@@ -0,0 +1,531 @@
...
+    # update intermediate output
+    output .= input
+
+    # perform pass over all remaining layers
+    backward!(input, Base.front(layers), output)
+
+end
+
+# when no layers are left, stop the inplace recursion
+backward!(input::Vector{T1}, layers::Tuple{}, output::Vector{T2}) where { T1 <: Real, T2 <: Real } = return
```
bvdmitri, line 242: Just again, better to replace `Vector` with `AbstractVector` everywhere.

src/nodes/flow/flow_models/flow_model.jl
```diff
@@ -0,0 +1,531 @@
...
+    # return result
+    return input
+    
+end
+
+# when calling backward, redirect to _backward
+backward(model::CompiledFlowModel, output::Vector{T}) where { T <: Real } = _backward(model, output)
+
+# for broadcasting over backward, fix the model for multiple inputs
+Broadcast.broadcasted(::typeof(backward), model::CompiledFlowModel, output::Vector{Vector{T}}) where { T <: Real } = broadcast(_backward, Ref(model), output)
```
bvdmitri, line 207: `AbstractVector{ <: AbstractVector{ <: Real } }`

src/nodes/flow/flow_models/flow_model.jl
```diff
@@ -0,0 +1,531 @@
...
+    dim    = getdim(model)
+    
+    # promote type for allocating output
+    T = promote_type(eltype(model), T1)
+
+    # allocate space for jacobian
+    J = zeros(T, dim, dim)
+    input = zeros(T, dim)
+    for k = 1:dim
+        J[k,k] = 1.0
```
bvdmitri, line 355: `one(T)` instead of `1.0`

src/nodes/flow/flow_models/flow_model.jl
```diff
@@ -0,0 +1,531 @@
...
+    
+    # fetch dimension
+    dim    = getdim(model)
+    
+    # promote type for allocating output
+    T = promote_type(eltype(model), T1, T2)
+    
+    # allocate space for input and jacobian
+    input = zeros(T, dim)
+    J .= 0.0
```
bvdmitri, line 475: same, `zero(T)`

src/nodes/flow/layers/additive_coupling_layer.jl
```diff
@@ -0,0 +1,294 @@
...
+```julia
+f = PlanarFlow()
+layer = AdditiveCouplingLayer(f)
+```
+
+This layer structure has been introduced in:
+
+Dinh, Laurent, David Krueger, and Yoshua Bengio. "Nice: Non-linear independent components estimation." _arXiv preprint_ arXiv:1410.8516 (2014).
+"""
+struct AdditiveCouplingLayer{ T <: NTuple{N,AbstractCouplingFlow} where { N } } <: AbstractCouplingLayer
```
bvdmitri, line 45: Interesting syntax, didn't know about that. Since I didn't know its possible I'm interested does it affect performance or not. Should be fine but we can discuss.

albertpod:
@bartvanerp @bvdmitri I think the `develop-flow` branch is very solid and should be merged.

src/nodes/flow/layers/additive_coupling_layer.jl
bartvanerp: The reason for this particular syntax was that it resulted in type-stable code.

src/nodes/flow/flow_models/flow_model.jl
```diff
@@ -0,0 +1,531 @@
...
+    
+    # fetch dimension
+    dim    = getdim(model)
+    
+    # promote type for allocating output
+    T = promote_type(eltype(model), T1, T2)
+    
+    # allocate space for output and jacobian
+    output = zeros(T, dim)
+    J .= 0.0
```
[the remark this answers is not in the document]
bartvanerp: Resolved by 92fa1f2

src/nodes/flow/flow_models/flow_model.jl
bartvanerp: Resolved by 92fa1f2

src/nodes/flow/flow_models/flow_model.jl
```diff
@@ -0,0 +1,531 @@
...
+    # promote type for allocating output
+    T = promote_type(eltype(model), T2)
+
+    # allocate space for intermediate output and jacobian
+    output_new = zeros(T, dim)
+    output_new .= output
+    J_new = zeros(T, dim, dim)
+    J_old = zeros(T, dim, dim)
+    for k = 1:dim
+        J_old[k,k] = 1.0
```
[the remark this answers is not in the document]
bartvanerp: Resolved by 92fa1f2

src/nodes/flow/flow_models/flow_model.jl
bartvanerp: Resolved by 92fa1f2

src/nodes/flow/flow_models/flow_model.jl
bartvanerp: Resolved by a3da751.

src/nodes/flow/flow_models/flow_model.jl
bartvanerp: Resolved by 0056d3a.

src/nodes/flow/flow_models/flow_model.jl
bartvanerp: Resolved by 0056d3a.

src/nodes/flow/flow_models/flow_model.jl
```diff
@@ -0,0 +1,531 @@
...
+    # calculate backward inverse jacobian over all layers
+    backward_inv_jacobian!(J, J_new, J_old, input, output_new, layers)
+
+end
+
+# inplace backward inverse jacobian calculation for a tuple of layers
+function backward_inv_jacobian!(J::Matrix{T1}, J_new::Matrix{T2}, J_old::Matrix{T3}, input::Vector{T4}, output_new::Vector{T5}, layers::T) where { T1 <: Real, T2 <: Real,  T3 <: Real, T4 <: Real, T5 <: Real, T <: NTuple{N,AbstractLayer} where N}
+
+    # check for specialized jacobians that are fixed (e.g. Permutation Matrix)
+    if typeof(last(layers)) <: PermutationLayer
```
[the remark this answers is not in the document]
bartvanerp: Hahah this is just stupid from my side. Good catch! Resolved by 7a0f784.

src/nodes/flow/coupling_flows/planar_flow.jl
bartvanerp: Resolved by 0056d3a.

src/nodes/flow/coupling_flows/planar_flow.jl
bartvanerp: Resolved by 0056d3a.

src/nodes/flow/coupling_flows/planar_flow.jl
bartvanerp: Resolved by the typing commits.

src/algebra/permutation_matrix.jl
bartvanerp: Resolved by 5801d21.

src/algebra/permutation_matrix.jl
bartvanerp: Resolved by 0e88df9.

src/algebra/flatten_tuple.jl
bartvanerp: Imported the flatten function from `TupleTools.jl`. Resolved by 29f246b.

src/algebra/helpers.jl
bartvanerp: Changed the `sigmoid` function to the `logistic` function from `StatsFuns`. Created an alias for easier usage. Apparently we already had (and still have) a `softmax` function in `ReactiveMP` in `math.jl`. However, this one clamps its output for inference purposes, but I am not sure whether this is desirable. Perhaps we should rename that one to `softmax_clamped` to prevent issues if we want to use the unclamped version of the softmax function. Partly resolved by 5ed8820.

src/algebra/permutation_matrix.jl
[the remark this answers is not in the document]
bartvanerp: Resolved by b98d86c and 4b18c08.

src/rules/flow/marginals.jl
```diff
@@ -0,0 +1,76 @@
...
+    # extract parameters of linearization
+    approximation = getapproximation(meta)
+    λ  = getλ(approximation)
+    L  = getL(approximation)
+    Wm = getWm(approximation)
+    Wc = getWc(approximation)
+
+    # calculate sigma points/vectors    
+    sqrtΣ = sqrt((L + λ)*Σ_out)
+    χ = Vector{Vector{Float64}}(undef, 2*L + 1)
```
bvdmitri, line 51: Can we change `Vector{Vector{Float64}}` to ``Vector{Vector{T}}` ?

bvdmitri (approved):

PR #49, opened 2021-09-29
HGF online structured

bvdmitri:
This PR extends HGF to support online structured variational inference when upper layer is a Gaussian signal.


ismailsenoz:
Below are couple of suggestions:

1.  Currently marginal computation rules do not take  the functional form of the messages into account. We can extend marginal computation rules such that the functional forms of the messages will be taken into account.
2. We can implement stochastic VI based methods to compute outgoing messages when the incoming messages are not Gaussian.
3. There are no checks for the numerical stability and this sometimes influence the performance.  We can prevent this by numerical tricks.

PR #63, opened 2021-11-08
Develop gmm pointmass

JimBeckers:
Implemented pointmass option for the mixture node.


bvdmitri (approved):
Nice catch @JimBeckers ! Once tests are passed we can merge.

PR #69, opened 2021-12-09
Optimization of ReactiveMP calculations

bartvanerp:
This (currently) draft pull request is meant to facilitate a discussion on the optimization of certain calculations in ReactiveMP and to act as a reference for later pull requests.

This pull request contains:
- Addition of a standard basis vector for allocation-free operations with a.o. the dot product node.
- Optimization of the `tr()` operator in the average energy calculation for the multivariate normal nodes.
- Refactoring of Flow update rules to use the present `prod()` functionality.
- Optimization of the `cholinv()` function, which halves the memory allocations when dealing with `BlasFloat`.
- Implementation of a joint `cholinv()` and `logdet()` function to speed up calculations.
- Optimization of Cholesky inverse times vector.
- Update of the `logdetcov` function used for entropy calculations by making use of the positive definiteness of covariance matrices.
- Perform `prod` with inplace operations to limit allocations.
- Refactor multiplication node output
- Extend addition and subtraction rules to use Distributions.jl `convolve` function by default.
- Forward all rules of the subtraction node to mirror the addition node.
- Reduced allocations in addition/subtraction marginal rules.
- Reduced allocations in mvnormal marginal rules.
- _This list will be updated over time..._

Possible elements that are still scheduled to appear in this PR are:
- Tests for the new standard basis vector.
- Tests for new joint `cholinv()` and `logdet()` function.
- Implementation of custom Wishart distribution to prevent automatic calculation of the normalization constant.
- Custom cholesky function that is fast by default and stable if necessary.
- Merge elementary factor nodes.
- Make marginal rules more efficient by not performing any unnecessary allocations.
- _This list will be updated over time..._

Below an explanation of the above topics will follow.


albertpod (commented):
Awesome contributions @bartvanerp!
Some of this stuff might be helpful for speeding up the AR node.
What do you think @bvdmitri?

bvdmitri:
Amazing work @bartvanerp! I'll review it later today.
I think this definitely can be used to speedup some computations in AR with new `StandardBasisVector`.

bvdmitri:
Very nice and clean PR Bart, great work. There are some minor issues but its fine. 

> Implementation of custom Wishart distribution to prevent automatic calculation of the normalization constant.

I definitely support this since I wanted to implement it super long ago 

I also wonder that it is better to sit together maybe and document all of this `non`-standard optimisation algebra tricks we can use in ReactiveMP. It is not a big problem though but I can imagine that someone who is unaware of lets say `StandardBasisVector ` will continue to use naive approach or someone who is unaware of `cholinv_logdet ` will use `logdet(cholinv())`. We do have a lot of cool algebra tricks in our package so we need to collect all of them in one single place in the documentation just for later reference.

bartvanerp:
## Addition of standard basis vector
This PR includes the `StandardBasisVector{len, ind}` object. It is a Cartesian standard basis vector of length `len` containing only zeros and only a single one at index `ind`. As an example, `StandardBasisVector{4,1}` simply represents `[1,0,0,0]`. 

This object prevents us from allocating an entire array of length `len`. Furthermore multiplications with this vector greatly simplify calculations. As an example consider the following case:
```julia
e1    = zeros(10)
e1[1] = 1
A     = randn(10,10)
x     = dot(e1, A, e1) # fast implementation of e1' * A * e1
```
Here the naive implementation `e1' *A * e1` allocates memory for the intermediate matrix-vector product and needs to do 110 multiplications and 99 additions (I think).
The better implementation using the dot product does not allocate anything, but performs 100 multiplications and 99 additions (I think).

The new implementation:
```julia
e1 = StandardBasisVector(10,1)
A  = randn(10,10)
x  = dot(e1, A, e1) 
```
simply returns `A[1,1]` without any calculations.

bartvanerp:
Indeed, let's bundle all comments in this pull request in a document later on, such that other people can benefit from this.

bartvanerp:
## Trace of more advanced matrix products
`ReactiveMP` currently already implements the `mul_trace(A, B)` function in `helpers.jl`, which calculates a fast trace over a matrix product `Tr(A*B)`, without allocating memory. To show how this works, we need to make use of the following identity [(source)](https://en.wikipedia.org/wiki/Trace_(linear_algebra))
image
This prevents us from performing a matrix-matrix product and its corresponding allocations.

This approach can also be extended to the matrix-outerproduct case `Tr(A*x*x')` as shown in [(source)](https://math.stackexchange.com/questions/293120/trace-of-a-matrix-times-outer-product), where we can rewrite
image
which prevents us from allocating two matrices (in the very naive case). Instead we just take the sum over some multiplications. 

Using both identities, we simplify the average energy calculations for the multivariate normal node from 
``` julia
tr(mean(q_Λ)*(v_out + v_mean + (m_out - m_mean)*(m_out - m_mean)'))
```
to
```julia
result = 0.0
    @turbo for k1 ∈ 1:dim, k2 ∈ 1:dim   # optimize trace operation (indices can be interchanges because of symmetry)
        result += S_Λ[k1,k2] * (V[k1,k2] + V[dim+k1,dim+k2] - V[dim+k1,k2] - V[k1,dim+k2] + (m[k1] - m[dim+k1]) * (m[k2] - m[dim+k2]))
    end
```
Here we also make use of the fact that our matrix is symmetrical and therefore we can interchange the indexing order for better memory accesses.

bartvanerp:
## Optimization of the cholesky inverse
The current implementation of `ReactiveMP` implements the cholesky inverse as 
```julia
inv(cholesky(PositiveFactorizations.Positive, Hermitian(x)))
```
Let's break this down. The part
```julia
cholesky(PositiveFactorizations.Positive, Hermitian(x))
```
is good. Calling `Hermitian()` prevents checks and speeds up computation time. Also `PositiveFactorizations.Postive` yields a performance improvement.
At this point we obtain a `Cholesky{T, Matrix{T}}` structure, which is our only source of allocation. Calling `inv` for `T <: LinearAlgebra.BlasFloat` then calls
```julia
inv!(C::Cholesky{<:BlasFloat,<:StridedMatrix}) =
    copytri!(LAPACK.potri!(C.uplo, C.factors), C.uplo, true)
inv(C::Cholesky{<:BlasFloat,<:StridedMatrix}) = inv!(copy(C))
```
https://github.com/JuliaLang/julia/blob/63f5b8a1654a77a91eccd619a665da5080ea6058/stdlib/LinearAlgebra/src/cholesky.jl#L688-L691
Which creates a copy on which we perform the inplace inverse. Our `cholinv()` function returns this copy and the original `Cholesky` structure is left for garbage collection.
This PR optimizes `cholinv()` to perform the inverse directly on the created `Cholesky` structure as 
```julia
function cholinv(x::AbstractMatrix{T}) where { T <: LinearAlgebra.BlasFloat }
    y = cholesky(PositiveFactorizations.Positive, Hermitian(x))
    LinearAlgebra.inv!(y)
    return y.factors
end
```
This about halves the memory usage

bartvanerp:
## Joint inverse and log-determinant
Some pieces of code in `ReactiveMP` use both an inverse and a log-determinant. Now first and foremost it is important to note that we are already calling `logdet()` instead of `log(det())` for numerical stability issues:
> "The determinant of a matrix is a high-degree polynomial, so it can be huge for even moderate-sized matrices. For example, the determinant of a Vandermonde matrix with consecutive integer elements increases super-factorially with the dimension of the matrix! For a 30 x 30 integer Vandermonde matrix, the determinant is too large to represent as a standard double-precision floating-point number." --  [(source)](https://blogs.sas.com/content/iml/2012/10/31/compute-the-log-determinant-of-a-matrix.html)

Although the `logdet()` function is stable, its calculation is quite expensive. When we are already calcuting an inverse it can be optimized using the intermediate result that we obtain using `cholinv`. Important here is that we are assuming that our matrix is positive definite. We make use of the following identity [(source)](https://math.stackexchange.com/questions/2001041/logarithm-of-the-determinant-of-a-positive-definite-matrix):
image
Luckily we obtain the cholesky decomposition already by calculating the inverse.

bartvanerp:
## Discussion points

### Use of `PositiveFactorizations.Positive`
For the cholesky decomposition, we pass the `PositiveFactorizations.Positive` keyword. On my end , however, this resulted into a slowdown of up to a factor 2 for low-dimensional matrices. Around a dimensionality of (50x50), adding the keyword gave similar performance as the default `cholesky` function. And for larger matrices it resulted in a speedup. Because we often deal with low-dimensional matrices, I have removed the keyword for now. 
Currently this change results into some test errors for the Wishart distribution and the Pointmass distribution:
`MethodError: no method matching cholesky(::LinearAlgebra.Hermitian{Float64, PDMats.PDMat{Float64, Matrix{Float64}}})`
However, this will be resolved once we implement our own `Wishart` distribution. For the pointmass distribution we try to calculate `cholinv` for a random matrix, which does not make sense in my opinion. https://github.com/biaslab/ReactiveMP.jl/blob/07391a67acd46a2ec591045f1971ad01f180ac69/test/distributions/test_pointmass.jl#L154
- _This list will grow over time..._

bvdmitri:
> ## Discussion points
> ### Use of `PositiveFactorizations.Positive`
> For the cholesky decomposition, we pass the `PositiveFactorizations.Positive` keyword. On my end , however, this resulted into a slowdown of up to a factor 2 for low-dimensional matrices. Around a dimensionality of (50x50), adding the keyword gave similar performance as the default `cholesky` function. And for larger matrices it resulted in a speedup. Because we often deal with low-dimensional matrices, I have removed the keyword for now. Currently this change results into some test errors for the Wishart distribution and the Pointmass distribution: `MethodError: no method matching cholesky(::LinearAlgebra.Hermitian{Float64, PDMats.PDMat{Float64, Matrix{Float64}}})` However, this will be resolved once we implement our own `Wishart` distribution. For the pointmass distribution we try to calculate `cholinv` for a random matrix, which does not make sense in my opinion.
> 
> https://github.com/biaslab/ReactiveMP.jl/blob/07391a67acd46a2ec591045f1971ad01f180ac69/test/distributions/test_pointmass.jl#L154
> 
> * _This list will grow over time..._

We cannot remove `PositiveFactorizations.Positive`. It deals properly with numerical instabilities and slightly non-posdef matrices. Julia's builtin implementation of `cholesky` is pretty weird when it comes to slightly non-posdef matrices, which may also happen even after `cholesky` itself.  Speed is always good but I'm in favour of computation correctness here - I would suggest to return `PositiveFactorizations.Positive` back.

bvdmitri:
Regarding `Positive` we can double check everything again of course, because 2x performance improvement at least worth to look into. But I remember trying with/without `Positive` and got numerical instabilities in very simple models. So IMO it is kinda dangerous to remove it now, but maybe we can also search for another similar library or maybe inspect code from `PositiveFactorisations` for possible performance improvements.

bartvanerp:
Could we instead leave this decision maybe to the user by using some global flag in reactive? For example:
image

bvdmitri:
Personally I do not like this kind of design with "safe"/"unsafe" functions, because you never know what situation is safe and what is not - so basically in 99% cases you would use safe version.

I mean we can get even better performance improvement if we would do all our computations in `Float16` for example. 😅But I think there should be a stop-line and possible numerical stability/accuracy is more important here. It is better not to tradeoff accuracy and performance with any matrix-related computation imo. It can blow-up pretty fast. 

If someone really wants to use "fast/unsafe" version of cholesky inverse in some message rules there is always generic `inv(cholseky(A))` available. But our `cholinv` should probably be accurate since we use it almost everywhere, even though it might be slower.

bvdmitri:
But I agree that 2x performance slowdown is pretty sad, as I said its worth to look into and maybe patch existing `Positive` implementation. It does not do anything super advance so such a slowdown is odd indeed.

bvdmitri:
> MethodError: no method matching cholesky(::LinearAlgebra.Hermitian{Float64, PDMats.PDMat{Float64, Matrix{Float64}}})

This error is useful actually. We definitely do some extra work here because based on this error it means that we compute a cholesky factorisation for `PDMat` object, but `PDMat` has already precomputed Cholesky factors afaik. I think the only one place is `Wishart` indeed and its better to implement our own version of `Wishart`.

One of the reasons why I didn't do that before is : what name to use? With `Normal` it was pretty easy - `NormalMeanVariance` etc. But `Wishart` itself is reserved and ideally we do not want to confuse our users with conflicting names from `Distributions.jl`. TBH I'm not sure how to resolve it rather than doing all computation in our hidden `Wishart2` structure but at the end convert to/return `Wishart` from `Distributions.jl`.

bartvanerp:
Depending on the outcome of whether we need `PositiveFactorizations`, I found another approach for calculating the cholesky inverse by directly calling the LAPACK library as:
```julia
function cholinv(A)
    B = LinearAlgebra.cholcopy(A)
    LinearAlgebra.LAPACK.potrf!('U', B)
    LinearAlgebra.LAPACK.potri!('U', B)
    return Hermitian(B)
end
```
This approach yields equal performance with respect to the default `cholesky()`. It might still require a check for positive definiteness as is currently present in `cholesky()`. The main benefit of this approach is that it uses half the memory. Naturally, we could extend this approach to make the cholesky inverse allocation free as:
```julia
function cholinv!(A)
    LinearAlgebra.LAPACK.potrf!('U', A)
    LinearAlgebra.LAPACK.potri!('U', A)
    return Hermitian(A)
end
```

bvdmitri:
@bartvanerp That is a very nice approach! However, we still need a generic version since LAPACK supports only `AbstractFloat` and does not play nicely with `ForwardDiff`. E.g.:

```julia
julia> L = rand(3, 3); A = L' * L;

julia> foo(A) = det(cholinv(reshape(A, (3, 3))))
foo (generic function with 1 method)

julia> ForwardDiff.gradient(foo, vec(A))
ERROR: MethodError: no method matching potrf!(::Char, ::Matrix{ForwardDiff.Dual{ForwardDiff.Tag{typeof(foo), Float64}, Float64, 9}})
```

I use similar approach with algebra helpers (for example `rank1update` uses BLAS) but it first checks if input arguments are supported by BLAS and otherwise fallback to a generic method that supports differentiation with ForwardDiff.

bvdmitri:
We can do the similar check I use for `rank1update` and use LAPACK for better efficiency. I wonder though - isnt standard builtin `cholesky` also uses LAPACK? I would expect it to do so...if not and if you versions yields the same result I would say it is a really good PR for Julia itself :)

bartvanerp:
Ah I just see that I used an old version of my code and the halving of memory was a result of using `inv` instead of `inv!`, apologies. However, it is possible to calculate a `cholinv` inplace, although this only applies to `BlasFloat`. This inplace cholesky inverse can also be written as 
```julia
function cholinv!(A)
    y = cholesky!(PositiveFactorizations.Positive, A)
    LinearAlgebra.inv!(y)
    return y.factors
end
```

bartvanerp:
I did some preliminary experiments regarding the `PositiveFactorizations` package. I created a (2x2) matrix `A` with a specific condition number. I used this matrix to generate a linear system as 
```julia
x = randn(2)
b = A*x
y = cholinv(A)*b
error[k] = norm(x - y)
```
and evaluated the error that resulted from this ill-conditioned inverse. For 1000 iterations per condition number I plotted the mean errors below, differentiating between using `PositiveFactorizations` or not and between taking the inverse directly (`inv`) or by backsubstitution (`sub`).
image
The x-axis shows the log-condition number. At some point the default cholesky decomposition fails, due to numerical instability, resulting into negative eigenvalues. The `PositiveFactorizations` package then still returns a matrix, because it takes the absolute value of the eigenvalues I believe, but also yields a higher error.

These results are very preliminary, but might give some intuition in its operations. I will further look into this and also explore why their approach outperforms the default `cholesky` function for large matrices.

bartvanerp:
If might also be worth the effort investigating the very basic cholesky algorithm of http://hua-zhou.github.io/teaching/biostatm280-2019spring/slides/10-chol/chol.html:
```julia
function cholesky(mati::AbstractMatrix)
    mat = copy(mati)
    nbrows, nbcols = size(mat)
    @inbounds for col=1:nbcols
        @inbounds for idx in 1:col-1
            mat[col, col] -= mat[col, idx]^2;
        end
        mat[col, col] = sqrt(mat[col, col])

        @inbounds for row in col+1: nbrows
            @inbounds for idx in 1:col-1
                mat[row, col] -= mat[row, idx]*mat[col, idx]
            end
            mat[row, col] = mat[row, col]/mat[col, col]
        end
    end
    return LowerTriangular(mat)
end
```
which yields similar stability performance in comparison to the `LinearAlgebra.cholesky` functions. It can also easily be used for inplace operations. Most importantly, this implementation demolishes the standard `LinearAlgebra.cholesky` implementation in terms of speed. I tested it for matrices of size, 2, 20, 200 and 2000 and even for (200x200) matrices this algorithm is faster than the conventional algorithm. The only difference is that this one outputs a `Triangular` matrix instead of a `Cholesky` object, something for which we could write a custom wrapper if we really want to.

bartvanerp:
I ran some more benchmarks for the different cholesky functions.
image
Of course, this is based on a single machine, but its results are still noteworthy.
The fast cholesky algorithm is the one of my previous comment and the blocked version is from https://arxiv.org/pdf/1812.02056.pdf. This version at some point switches to blocked matrix multiplications to make use of the multithreaded blas library instead of performing calculations on a single thread. This introduces some overhead, but is worth it in the long run. (The timings of the fast cholesky function blew up, so it is not plotted on the full domain.)

albertpod:
Great work guys. May I throw two cents in?
I am currently experiencing a huge degrade in performance (over 100X) when working with NARMAX model that uses a Flow node.
Maybe some of you may know what's the heck?

bvdmitri (approved):
Great job @bartvanerp !

PR #70, opened 2021-12-10
Marginal constraint

albertpod:
Here is a small PR that enables immobilizing the marginal distribution such that it remains fixed during inference.


albertpod:
I personally find this PR a little obscure without an example. Should we add it to the description? Providing a separate demo for the constrain seems like overkill.

bartvanerp:
Looks good to me. I think it would be useful to extend the `API_tutorial.ipynb` to include an example with this form constraint.

bvdmitri (approved):

PR #81, opened 2022-01-19
Add GammaMM notebook

albertpod:
This PR adds an example of inference in the Gamma Mixture model.


bartvanerp (approved):

bvdmitri:
`as_constraint` and `bs_constraint` variables in the model specification are not used - it is probably better to remove them. Other than that IMO is ready to merge.

bartvanerp (approved):

PR #86, opened 2022-02-14
Add mean-field rules for AR node

albertpod:
This PR extends the AR node by supplying rules for mean-field factorization around the node.


bvdmitri (commented):
Looks very good to me. P.S. I did not check the actual correctness of the rules. Probably needs a second pair of eyes to double-check.

src/rules/autoregressive/marginals.jl
```diff
@@ -1,9 +1,9 @@
 
-@marginalrule AR(:y_x) (m_y::NormalDistributionsFamily, m_x::NormalDistributionsFamily, q_θ::NormalDistributionsFamily, q_γ::GammaShapeRate, meta::ARMeta) = begin
+@marginalrule AR(:y_x) (m_y::Any, m_x::Any, q_θ::NormalDistributionsFamily, q_γ::Any, meta::ARMeta) = begin
```
bvdmitri, line 2: I'm hesitant of using `::Any` for messages, for marginals it is fine. We can discuss.

albertpod:
I've updated the rules.

PR #97, opened 2022-04-25
Fix :var messages

### LINKED ISSUE #96: Outgoing BP rules for :var interface of NormalMeanVariance node.
### Discussed in https://github.com/biaslab/ReactiveMP.jl/discussions/95

Hi guys,
(@ismailsenoz @semihakbayrak @bvdmitri)
How sure we are about the BP message for `:var` interface that is associated with variance?

We had a look a few days ago with @bartvanerp and it seemed to us that the normalization of the message is incorrect.
I see there are tests for these rules, but I am still concerned about some missing factors in:
```julia
ContinuousUnivariateLogPdf(DomainSets.HalfLine(), (x) -> -log(m_μ_cov + x) / 2 - 1 / (2x) * (m_out_mean - m_μ_mean) ^ 2)
```

There is an issue with BP message from `GaussianMeanVariance` node. The correct version when both `:mean` and `:out` inbounds messages are Gaussians should be:
```
(x) -> -log(m_μ_cov + m_out_cov + x) / 2 - 1 / (2*(m_μ_cov + m_out_cov + x)) * (m_out_mean - m_μ_mean) ^ 2)
```

bvdmitri:
Fixes #96. @albertpod, @bartvanerp, @ismailsenoz please double check.

### DIFF
--- a/src/rules/normal_mean_variance/var.jl
+++ b/src/rules/normal_mean_variance/var.jl
@@ -5,19 +5,19 @@
     m_out_mean        = mean(m_out)
     m_μ_mean, m_μ_cov = mean_cov(m_μ)
     
-    return ContinuousUnivariateLogPdf(DomainSets.HalfLine(), (x) -> -log(m_μ_cov + x) / 2 - 1 / (2x) * (m_out_mean - m_μ_mean) ^ 2)
+    return ContinuousUnivariateLogPdf(DomainSets.HalfLine(), (x) -> -log(m_μ_cov + x) / 2 - 1 / (2*(x + m_μ_cov)) * (m_out_mean - m_μ_mean) ^ 2)
 end
 
 @rule NormalMeanVariance(:v, Marginalisation) (m_out::UnivariateNormalDistributionsFamily, m_μ::PointMass) = begin 
     m_out_mean, m_out_cov = mean_cov(m_out)
     m_μ_mean              = mean(m_μ)
 
-    return ContinuousUnivariateLogPdf(DomainSets.HalfLine(), (x) -> -log(m_out_cov + x) / 2 - 1 / (2x) * (m_out_mean - m_μ_mean) ^ 2)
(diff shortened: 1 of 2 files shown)


src/rules/normal_mean_variance/var.jl
```diff
@@ -5,19 +5,19 @@
...
-    return ContinuousUnivariateLogPdf(DomainSets.HalfLine(), (x) -> -log(m_out_cov + x) / 2 - 1 / (2x) * (m_out_mean - m_μ_mean) ^ 2)
+    return ContinuousUnivariateLogPdf(DomainSets.HalfLine(), (x) -> -log(m_out_cov + x) / 2 - 1 / (2*(x + m_out_cov)) * (m_out_mean - m_μ_mean) ^ 2)
 end
 
 @rule NormalMeanVariance(:v, Marginalisation) (m_out::UnivariateNormalDistributionsFamily, m_μ::UnivariateNormalDistributionsFamily) = begin 
     m_out_mean, m_out_cov = mean_cov(m_out)
     m_μ_mean, m_μ_cov     = mean_cov(m_μ)
 
-    return ContinuousUnivariateLogPdf(DomainSets.HalfLine(), (x) -> -log(m_out_cov + m_μ_cov + x) / 2 - 1 / (2x) * (m_out_mean - m_μ_mean) ^ 2)
+    return ContinuousUnivariateLogPdf(DomainSets.HalfLine(), (x) -> -log(m_out_cov + m_μ_cov + x) / 2 - 1 / (2*(m_out_cov + m_μ_cov + x)) * (m_out_mean - m_μ_mean) ^ 2)
```
bartvanerp, line 22: Do we have derivations for this one? I am not sure whether an extension is so straightforward.

bvdmitri: image

bartvanerp (approved):

PR #125, opened 2022-06-01
docs: add list of exported methods #114

### LINKED ISSUE #114: Docs should contain list of exported methods
Feedback from users:

It is bad practice (common to many programming languages) to import all methods implicitly, e.g., `from numpy import *`, as it might silently overwrite a reference. So, users would like to explicitly `import ReactiveMP: update!, ... ` instead of `using ReactiveMP`. 

However, the documentation does not list all methods that ReactiveMP exports. Adding the remainder would be helpful.

bvdmitri:
This PR adds the list of exported methods to the documentation. Fixes #114

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -33,7 +33,8 @@ makedocs(
                 "Flow"     => "lib/nodes/flow.md"
             ],
             "Math utils"       => "lib/math.md",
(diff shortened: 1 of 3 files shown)


wmkouw (approved):
Looks great, Dmitry! Thanks.

PR #126, opened 2022-06-02
Add mode calculation for GammaShapeRate distribution

wmkouw:
The mode of a GammaShapeRate distribution is now calculated by extracting the shape and rate parameters, inverting the rate to a scale parameter, defining a GammaShapeScale distribution and calling Distributions.jl's mode calculation.

Notes:
-  GammaShapeScale inherits mode calculation because it is a redefinition of Distributions.Gamma.
- In case of shape < 1, mode calculation throws identical error to Distributions.Gamma mode calculation.

### DIFF
--- a/src/distributions/gamma.jl
+++ b/src/distributions/gamma.jl
@@ -9,8 +9,6 @@ const GammaDistributionsFamily{T} = Union{GammaShapeScale{T}, GammaShapeRate{T}}
 
 Distributions.cov(dist::GammaDistributionsFamily) = var(dist)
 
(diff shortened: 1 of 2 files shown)


wmkouw:
Forgot to run `make format`.

2nd and 3d commit updated PR with:
1. Moved mode function from `gamma.jl` to `gamma_shape_rate.jl`.
2. Broke line twice to prevent overshooting margin.

bvdmitri (approved):
Thanks @wmkouw . Nice catch!

Issue #129, opened 2022-06-13
Add Poisson node


albertpod:
`ReactiveMP.jl` is missing Poisson node.

The node exists in [ForneyLab.jl](https://github.com/biaslab/ForneyLab.jl/blob/master/src/factor_nodes/poisson.jl).

Any volunteers to port?

bvdmitri:
Fixed with #132

PR #132, opened 2022-06-13
Add Poisson node

albertpod:
This PR addresses issue #129


src/rules/poisson/out.jl
```diff
@@ -0,0 +1,5 @@
+export rule
+
+@rule Poisson(:out, Marginalisation) (m_l::PointMass,) = Poisson(mean(m_l))
+
+@rule Poisson(:out, Marginalisation) (q_l::GammaDistributionsFamily,) = Poisson((shape(q_l) - 1 // 2) / rate(q_l))
```
bartvanerp, line 5: I end up with a digamma function here: `lambda = exp(digamma(alpha))/beta`. Also saw that there are some approximations to this term: https://en.wikipedia.org/wiki/Digamma_function#Computation_and_approximation. The question is whether we always want to use this approximation.

src/nodes/poisson.jl
```diff
@@ -0,0 +1,29 @@
...
+        end
+        return s
+    else
+        # Try `BigFloat`
+        return convert(T, __approximate_powersum(BigFloat, l, 150))
+    end
+end
+
+@average_energy Poisson (q_out::Any, q_l::Any) =
+    mean(q_l) - mean(q_out) * mean(log, q_l) + exp(-mean(q_out)) * __approximate_powersum(Float64, mean(q_out))
```
bartvanerp, line 29: Derivation seems correct. I am not really up to speed of the approximation of `E_q(x)[ln(x!)]`, but could we also get away with the (perhaps simpler) Stirling's approximation? Furthermore I think we want to add some specialization for this average energy function, such that we do not always compute this approximation.

albertpod: Will it actually help?

bvdmitri: I think you still need to approximate this power sum even if `q_out` is a `PointMass`. We've spent quite some time yesterday to figure out how we can further simplify this expression. The computation is super unstable for large values. Distributions.jl has an interesting approximation here: https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariate/discrete/poisson.jl#L68, but I'm not sure how to incorporate it in our case (it should be possible, maybe I will spent more time on this). They do compute a slightly different quantity though. Any cool ideas are welcome!

bvdmitri: Ah, well, I understood you wrong. You are correct @bartvanerp if `Poisson` used for observations. The original approximation, however, assumes Poisson as a prior, not as a likelihood for your observations. So we need it anyway I believe. But it seems reasonable to add an extra dispatch method where `q_out` is a `PointMass`, it simplifies the computation indeed.

albertpod:
Please look up the fixes

bartvanerp (approved):

bvdmitri (approved):

PR #147, opened 2022-06-21
Dev broadcasting syntax v2

bvdmitri:
Support for broadcasting syntax in the model specification from `GraphPPL.jl`


bvdmitri:
Tests are failing because GraphPPL.jl "v2.1.0" is not available in CI. Tests are passing locally.

Issue #148, opened 2022-06-23
average energy computation for the Wishart node


albertpod:
I've been looking into the computation of average energy for the `Inverse-Wishart `node.
So I had a look at the computation of this quantity for `Wishart` node as well.
It appears to me that the average energy computation for the `Wishart` node is incorrect, i.e.,

`@average_energy Wishart (q_out::Any, q_ν::Any, q_S::Any)`

You would normally expect `q_out` to be Wishart distribution, meaning that we'd need to take the log expectation of `q_out`, not just `logdet(m_q_out)` where `m_q_out = mean(q_out)`. 

It seems to me that [ForneyLab.jl ](https://github.com/biaslab/ForneyLab.jl/blob/fca07c3bcb3a5abe7393dfb275a2e47e54fd68aa/src/factor_nodes/wishart.jl#L137)does it properly. Is there something I miss?

bartvanerp:
I think you might be right. Previously (before @bvdmitri introduced the `mean(logdet, x)` syntax) this also was incorrect for the `Gamma` node. It seems that we overlooked this one.

albertpod:
I guess we are missing the `mean(inverse, x) `function for some distributions.

bvdmitri:
Indeed it looks like an artefact of the old confusion between `unsafeDetLogMean(...)` and `logdet(mean(...))`, which was wrong. We replaced `logdet(mean())` with `mean(logdet, ...)` but overlooked this one. Thanks @albertpod !

bvdmitri:
Fixed with #149 . Thanks @albertpod !

PR #149, opened 2022-06-24
Fix average energy for Wishart

albertpod:
This PR fixes issue #148.

### DIFF
--- a/src/distributions/wishart.jl
+++ b/src/distributions/wishart.jl
@@ -11,6 +11,12 @@ function Distributions.mean(::typeof(logdet), distribution::Wishart)
     return mapreduce(i -> digamma((ν + 1 - i) / 2), +, 1:d) + d * log(2) + logdet(S)
 end
 
+function Distributions.mean(::typeof(cholinv), distribution::Wishart)
+    d    = ndims(distribution)
+    ν, S = params(distribution)
+    sum([digamma.(0.5*(ν + 1 - i)) for i = 1:d]) + d*log(2) + logdet(S)
+end
+
 vague(::Type{<:Wishart}, dims::Int) = Wishart(dims, Matrix(Diagonal(huge .* ones(dims))))
 
 Base.ndims(dist::Wishart) = Distributions.dim(dist)
--- a/src/distributions/wishart.jl
+++ b/src/distributions/wishart.jl
@@ -11,12 +11,6 @@ function Distributions.mean(::typeof(logdet), distribution::Wishart)
     return mapreduce(i -> digamma((ν + 1 - i) / 2), +, 1:d) + d * log(2) + logdet(S)
 end
 
-function Distributions.mean(::typeof(cholinv), distribution::Wishart)
-    d    = ndims(distribution)
-    ν, S = params(distribution)
-    sum([digamma.(0.5 * (ν + 1 - i)) for i in 1:d]) + d * log(2) + logdet(S)
-end
-
 vague(::Type{<:Wishart}, dims::Int) = Wishart(dims, Matrix(Diagonal(huge .* ones(dims))))
 
 Base.ndims(dist::Wishart) = Distributions.dim(dist)
--- a/src/distributions/wishart.jl
+++ b/src/distributions/wishart.jl
@@ -14,7 +14,7 @@ end
 function Distributions.mean(::typeof(cholinv), distribution::Wishart)
     d    = ndims(distribution)
     ν, S = params(distribution)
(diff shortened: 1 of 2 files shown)


src/nodes/wishart.jl
```diff
@@ -13,8 +13,8 @@ import Distributions: Wishart
     m_q_out = mean(q_out)
 
     return 0.5 * (
-        m_q_ν * (logdet(m_q_S) + d * log(2)) -
-        logdet(m_q_out) * (m_q_ν - d - 1.0) +
-        tr(cholinv(m_q_S) * m_q_out) + 0.5 * d * (d - 1) * logπ
+        m_q_ν * (mean(logdet, q_S) + d * log(2)) -
+        mean(logdet, q_out) * (m_q_ν - d - 1.0) +
+        tr(mean(cholinv, q_S) * m_q_out) + 0.5 * d * (d - 1) * logπ
     ) + mapreduce(i -> loggamma(0.5 * (m_q_ν + 1.0 - i)), +, 1:d)
```
bartvanerp, line 19: Are we allowed to do this? I see that it is the case in ForneyLab. So basically we compute `E_q(v)[lnGamma(v/2 - (j-1)/2)]` if I am not mistaken. Can we state `E_q(v)[lnGamma(v/2 - (j-1)/2)] = lnGamma(E_q(v)[v]/2 - (j-1)/2)`?

src/distributions/wishart.jl
```diff
@@ -11,6 +11,12 @@ function Distributions.mean(::typeof(logdet), distribution::Wishart)
     return mapreduce(i -> digamma((ν + 1 - i) / 2), +, 1:d) + d * log(2) + logdet(S)
 end
 
+function Distributions.mean(::typeof(cholinv), distribution::Wishart)
```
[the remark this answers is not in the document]
bvdmitri: Yes, @bartvanerp is right. The method definition should be `function Distribution.mean(::typeof(inv), ::Wishart)` in the same way as for [`PointMass`](https://github.com/biaslab/ReactiveMP.jl/blob/754e15c1b8b509ba52f337e5836a3aff5556a1a0/src/distributions/pointmass.jl#L92).

src/nodes/wishart.jl
```diff
@@ -13,8 +13,8 @@ import Distributions: Wishart
     m_q_out = mean(q_out)
 
     return 0.5 * (
-        m_q_ν * (logdet(m_q_S) + d * log(2)) -
-        logdet(m_q_out) * (m_q_ν - d - 1.0) +
-        tr(cholinv(m_q_S) * m_q_out) + 0.5 * d * (d - 1) * logπ
+        m_q_ν * (mean(logdet, q_S) + d * log(2)) -
+        mean(logdet, q_out) * (m_q_ν - d - 1.0) +
+        tr(mean(cholinv, q_S) * m_q_out) + 0.5 * d * (d - 1) * logπ
```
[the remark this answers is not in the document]
bvdmitri: @bartvanerp, @albertpod  `mean(inv, dist)`

src/distributions/wishart.jl
```diff
@@ -11,6 +11,12 @@ function Distributions.mean(::typeof(logdet), distribution::Wishart)
...
+    d    = ndims(distribution)
+    ν, S = params(distribution)
+    sum([digamma.(0.5 * (ν + 1 - i)) for i in 1:d]) + d * log(2) + logdet(S)
```
bvdmitri, line 17: Replace `sum([ ... ])` with `sum(...)`. Note that I just removed squared brackets:
```julia
julia> foo(n) = sum([ i for i in 1:n ])
foo (generic function with 1 method)

julia> bar(n) = sum(i for i in 1:n)
bar (generic function with 1 method)

julia> @btime foo(100)
  97.671 ns (1 allocation: 896 bytes)
5050

julia> @btime bar(100)
  1.514 ns (0 allocations: 0 bytes)
5050
```

I also not sure about `.` in `digamma.`. It is a bit confusing, I suppose this code directly ported from ForneyLab. Probably we can simply omit `.`?

bartvanerp: Wow didn't know this

bvdmitri: @bartvanerp The reason is this :) 
```julia
julia> typeof((i for i in 1:10))
Base.Generator{UnitRange{Int64}, typeof(identity)}

julia> typeof([i for i in 1:10])
Vector{Int64}
```
So "bracket-free" version does not allocate and generates items lazily.

bvdmitri (approved):

bartvanerp (approved):

PR #154, opened 2022-06-29
feat: add diagnostic checks for Bethe Free Energy stream

### LINKED ISSUE #136: Report where free energy computes `NaN`
`inference` function should check and report at what node or edge Free Energy computation fails with `NaN`

bvdmitri:
This PR fixes #136


bartvanerp (approved):
Great feature! I think it looks good, but have not tried it on my machine. Perhaps we could add some tests?

bvdmitri:
@bartvanerp You are right. I added tests for all cases, found one ctrl-c ctrl-v bug 🙈

PR #160, opened 2022-07-04
feat: new functional dependencies pipeline `RequireInboundMarginal`

bvdmitri:
This PR adds `RequireInboundMarginalFunctionalDependencies`, which specifies that in order to compute a message on some edge rule requires marginal on the same edge. The interface is pretty much the same as for `RequireInboundFunctionalDependencies`.

Do not merge, tests are not implemented (yet).


MagnusKoudahl (approved):
Updated some doc strings, looks good to me otherwise

PR #161, opened 2022-07-04
Add Inverse-Wishart

albertpod:
This PR contributes to `ReactiveMP.jl` in 3 aspects:

1. Adds `InvWishart` distribution. (`InverseWishart` from [Distributions.jl](https://github.com/JuliaStats/Distributions.jl/issues/1569) won't work as we need to pass improper distributions through the graph). `InvWishart` implements `prod`, `entropy`, etc.
2. Adds `InvWishart`node.
3. Adds variational update rules to compute the posterior covariance matrix of `MvNormalMeanCovariance` node.


src/distributions/wishart.jl
```diff
@@ -11,6 +11,11 @@ function Distributions.mean(::typeof(logdet), distribution::Wishart)
     return mapreduce(i -> digamma((ν + 1 - i) / 2), +, 1:d) + d * log(2) + logdet(S)
 end
 
+function Distributions.mean(::typeof(inv), distribution::Wishart)
+    ν, S = params(distribution)
+    return ν*inv(S)
```
bvdmitri, line 16: `cholinv` instead of `inv` for positive-definite matrices is better.

src/rules/mv_normal_mean_covariance/covariance.jl
```diff
@@ -0,0 +1,24 @@
+
+# Variational                       # 
+# --------------------------------- #
+@rule MvNormalMeanCovariance(:Σ, Marginalisation) (q_out::Any, q_μ::Any) = begin
+    m_out, v_out   = mean_cov(q_out)
+    m_mean, v_mean = mean_cov(q_μ)
+
+    df = ndims(q_μ)
+    S  = Matrix(Hermitian(v_mean + v_out + (m_mean - m_out) * (m_mean - m_out)'))
```
bvdmitri, line 9: `Matrix(Hermitian(...))` is redundant for `InvWishart`

src/distributions/inv_wishart.jl
```diff
@@ -0,0 +1,74 @@
...
+Distributions.var(dist::InvWishart)  = var(InverseWishart(params(dist)...))
+Distributions.cov(dist::InvWishart)  = cov(InverseWishart(params(dist)...))
+Distributions.mode(dist::InvWishart) = mode(InverseWishart(params(dist)...))
+Distributions.dim(dist::InvWishart)  = size(dist.S, 1)
+
+# from "Parametric Bayesian Estimation of Differential Entropy and Relative Entropy" Gupta et al.
+function Distributions.entropy(dist::InvWishart)
+    d = dim(dist)
+    ν, S = params(dist)
+    0.25 * d * (d - 1) * logπ + mapreduce(i -> loggamma(0.5 * (ν + 1.0 - i)), +, 1:d) +
```
bvdmitri, line 33: I would suggest to change all `Float64` literals to their `Int` (e.g. instead of `0.5 * a` do `a / 2`). You can read about the reasoning here: [Avoid using floats for numeric literals in generic code when possible](https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-using-floats-for-numeric-literals-in-generic-code-when-possible)

src/distributions/inv_wishart.jl
```diff
@@ -0,0 +1,74 @@
...
+    ν, S = params(dist)
+    return -(mapreduce(i -> digamma((ν + 1 - i) / 2), +, 1:d) + d * log(2) - logdet(S))
+end
+
+function Distributions.mean(::typeof(inv), dist::InvWishart)
+    ν, S = params(dist)
+    return ν * cholinv(S)
+end
+
+vague(::Type{<:InvWishart}, dims::Integer) = InvWishart(dims, inv(Matrix(Diagonal(huge .* ones(dims)))))
```
bvdmitri, line 49:
```suggestion
vague(::Type{<:InvWishart}, dims::Integer) = InvWishart(dims, tiny .* diageye(dims))
```

src/distributions/inv_wishart.jl
```diff
@@ -0,0 +1,74 @@
...
+
+prod_analytical_rule(::Type{<:InvWishart}, ::Type{<:InvWishart}) = ProdAnalyticalRuleAvailable()
+
+function prod(::ProdAnalytical, left::InvWishart, right::InvWishart)
+    d = dim(left)
+
+    ldf, lS = params(left)
+    rdf, rS = params(right)
+
+    V = lS + rS |> Hermitian |> Matrix
```
bvdmitri, line 69:
```suggestion
    V = lS + rS 
```

src/nodes/inv_wishart.jl
```diff
@@ -0,0 +1,18 @@
...
+import Distributions: InverseWishart
+
+@node InvWishart Stochastic [out, (ν, aliases = [df]), (S, aliases = [scale])]
+
+@average_energy InvWishart (q_out::Any, q_ν::PointMass, q_S::Any) = begin
+    d = dim(q_out)
+
+    m_q_ν = mean(q_ν)
+
+    return 0.5 * (
```
bvdmitri, line 13: Same comment here: [Avoid using floats for numeric literals in generic code when possible](https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-using-floats-for-numeric-literals-in-generic-code-when-possible)

src/rules/mv_normal_mean_covariance/covariance.jl
```diff
@@ -0,0 +1,24 @@
...
+end
+
+@rule MvNormalMeanCovariance(:Σ, Marginalisation) (q_out_μ::Any,) = begin
+    m_out_μ, v_out_μ = mean_cov(q_out_μ)
+
+    d = div(ndims(q_out_μ), 2)
+
+    mdiff = @views m_out_μ[1:d] - m_out_μ[d+1:end]
+    vdiff = @views v_out_μ[1:d, 1:d] - v_out_μ[1:d, d+1:end] - v_out_μ[d+1:end, 1:d] + v_out_μ[d+1:end, d+1:end]
+    S     = Matrix(Hermitian(vdiff + mdiff * mdiff'))
```
bvdmitri, line 21:
```suggestion
    S     = vdiff + mdiff * mdiff')
```

src/distributions/inv_wishart.jl
```diff
@@ -0,0 +1,74 @@
...
+end
+
+function Base.convert(::Type{Wishart}, dist::InvWishart)
+    ν, S = params(dist)
+    return InverseWishart(ν, cholinv(S))
+end
+
+prod_analytical_rule(::Type{<:InvWishart}, ::Type{<:InvWishart}) = ProdAnalyticalRuleAvailable()
+
+function prod(::ProdAnalytical, left::InvWishart, right::InvWishart)
```
bvdmitri, line 63: What would happen if I assign `InverseWishart` from `Distribution.jl`. Does it mean that we always have to use our own structure? I have an idea how we can circumvent this, we can discuss in person.

src/distributions/wishart.jl
albertpod: fixed

src/rules/mv_normal_mean_covariance/covariance.jl
albertpod: Theoretically, yes. I wonder if that may not be the case during the actual computations.

Are you sure we can skip this step? We do the same thing for `rules/mv_norma_mean_precision/precision.jl`

src/distributions/inv_wishart.jl
ismailsenoz: conversion between inverse Wishart and Wishart is not correct, unfortunately.  if Sigma with dimension n is distributed according to Wishart with nu degrees of freedom and matrix K, then inverse Sigma is distributed Inverse Wishart with nu-n+1 degree of freedom with matrix inverse K .

albertpod: I agree that convert is incorrect. This method should be removed

However, I dont think that what you say is right, @ismailsenoz

If X~W(n,S) then X^(-1)~IW(n, S^(-1))

ismailsenoz: Please see https://www.cs.cmu.edu/~andrewgw/tprocess.pdf for the definition of inverse Wishart and its conversion to Wishart. It might be due to the parameterization difference, but the parameterization in the ref has advantages.

ismailsenoz: So apparently, there is a diversion indeed. Maybe, we should discuss which parameterization we want to use?

bvdmitri: I'm fine with the current parametrisation in this PR as it directly translates to the parametrisation in `InverseWishart` from `Distributions.jl`.

albertpod:
I haven't come up with a good test for the expectations of `logdet` and `inv`.

The other stuff is ready for review.

bvdmitri:
Great work @albertpod . I will take over this PR and will try to fix `with_float_conversions` argument. Ideally it should pass with `true`, but if not possible its also fine.

bvdmitri:
I fixed `with_float_conversions = true` in tests. As soon as CI is passing we can merge.

bvdmitri (approved):

PR #163, opened 2022-07-11
Wishart/InverseWishart messages

bvdmitri:
This PR implements `WishartMessage` in addition to `InvWishart` (renamed to `InverseWishartMessage`). These structures are fast to create, do not check their input arguments and are way faster then their equivalents from `Distributions.jl`. 

All the complex work has been done in `inverse-wishart-node` branch, this PR is only about small refactoring, better efficiency and performance. However, there are other changes in this PR:

- Users are free to use `InverseWishart` in the `@model` specification, `InverseWishartMessage` (`InvWishart`) is a private structure (we may consider something similar for `MvNormal` as @bartvanerp  suggested)
- I renamed `inv_wishart.*` files to `wishart_inverse.*` filenames, so `wishart.*` and `wishart_inverse.*` are located close to each other in IDE (sorry for broken `git diff`).
- The overall performance is way better for the `Wishart` case.

For benchmarking I wrote a simple IID model with 1500 observations, when I started the benchmark result were 
- Wishart 159.850 ms (1434661 allocations: 93.97 MiB)
- InvWishart 129.347 ms (1374725 allocations: 82.57 MiB)

After refactoring:
- Wishart 115.419 ms (1269698 allocations: 82.53 MiB)
- InverseWishart 111.750 ms (1254708 allocations: 79.28 MiB)


albertpod:
Quite some tests involving Wishart priors fail.

bvdmitri:
@albertpod I'll take a look into it

albertpod (approved):

Issue #120, opened 2022-05-20
Support of message passing for Boolean functions


albertpod:
Currently, `ReactiveMP.jl` does not support inference in Belief networks composed of Bernoulli random variables (RVs).

One good example of such a network is shown in [Chapter 2](https://www.mbmlbook.com/LearningSkills.html) of Bishop's **Model-Based Machine Learning** book.

The dependencies between these variables are enforced by logic functions, e.g., `AND`, `NOT`, `OR`, etc.

`ForneyLab.jl `has the demo on [building `AND` node](https://github.com/biaslab/ForneyLab.jl/blob/master/demo/implementing_additional_nodes.ipynb). 

A related [issue](https://github.com/biaslab/ForneyLab.jl/issues/200) was initially opened in `ForneyLab.jl`.

bvdmitri:
Fixed in #170

PR #170, opened 2022-07-20
Support of message passing for Boolean functions

albertpod:
This PR addresses issue #120.

Thanks to @Chengfeng-Jia and @Sepideh-Adamiat for deriving rules.


albertpod:
This PR misses a nice demo demonstrating these functions' applicability.

Once the demo is finished, we can promote this PR.

test/runtests.jl
```diff
@@ -150,8 +150,12 @@ end
     addtests("nodes/test_mv_normal_mean_covariance.jl")
     addtests("nodes/test_poisson.jl")
     addtests("nodes/test_wishart_inverse.jl")
+    addtests("nodes/test_or.jl")
+    addtests("nodes/test_not.jl")
 
     addtests("rules/uniform/test_out.jl")
+    addtests("nodes/test_and.jl")
+    addtests("nodes/test_implication.jl")
```
bartvanerp, line 158: move these tests to the `nodes` section.

src/nodes/not.jl
```diff
@@ -0,0 +1,5 @@
+export NOT
+
+struct NOT end
+
+@node NOT Deterministic [out, in1]
```
bartvanerp, line 5: rename to `in`

bartvanerp: And adjust in future parts of the code (e.g. rules and tests)

src/rules/or/out.jl
```diff
@@ -0,0 +1,8 @@
+@rule OR(:out, Marginalisation) (
+    m_in1::Bernoulli,
+    m_in2::Bernoulli
+) = begin
+    pin1, pin2 = mean(m_in1), mean(m_in2)
+
+    return Bernoulli(pin1 + pin2 - pin1 * pin2)
```
bartvanerp, line 7: Please double check this one, I end up with `pin1 + pin2 - 2 * pin1 * pin2`

src/rules/and/marginals.jl
```diff
@@ -0,0 +1,8 @@
+@marginalrule AND(:in1_in2) (
+    m_out::Bernoulli,
+    m_in1::Bernoulli,
+    m_in2::Bernoulli
+) = begin
+    pin1, pin2, pout = mean(m_in1), mean(m_in2), mean(m_out)
+    return Contingency([(1-pin1)*(1-pin2)*(1-pout) (1-pin1)*pin2*(1-pout); pin1*(1-pin2)*(1-pout) pin1*pin2*pout])
```
bartvanerp, line 7: The `Contingency` object is not automatically normalized. I think we want to extend the `contingency.jl` to do so, otherwise users will get tables, which are unnormalized, whereas they expect them to be normalized.

bartvanerp, line 7: Furthermore, this `contingency` object has two dimensions and is asymmetric. As a user I would like to know how to interpret this table and therefore require something like axes labels. Perhaps an extension of the `Contingency` object is required for this purpose.

src/rules/or/out.jl
ismailsenoz: I ended up with ` pin1 + pin2 - pin1 * pin2 ` as well.  @bartvanerp Could it be that you omitted 'in1 = 1, in2=1'?

src/nodes/implication.jl
```diff
@@ -0,0 +1,5 @@
+export IMPL
+
+struct IMPL end
+
+@node IMPL Deterministic [out, in1, in2]
```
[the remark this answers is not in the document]
ismailsenoz: I think @bartvanerp's comment is critical and should be addressed. A possible fix would include the truth tables for all logical operations.

src/rules/or/out.jl
bartvanerp: Ah yeah you are absolutely right, accidentally did the computations for a XOR operation

src/nodes/implication.jl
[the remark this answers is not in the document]
albertpod: Perhaps good to add docstrings for every node. The implication is as "famous" as `OR` or `AND` operators.

src/nodes/and.jl
```diff
@@ -0,0 +1,5 @@
+export AND
+
+struct AND end
+
+@node AND Deterministic [out, in1, in2]
```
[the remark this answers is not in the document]
albertpod: This is a good suggestion, although `&` isn't the correct alias in my opinion as it refers to bitwise AND, not to logical AND (`&&`). Unlike `&`, `&&` is not a function but an operator. I couldn't figure out how to overload the operator.

src/ReactiveMP.jl
```diff
@@ -128,11 +128,15 @@ include("nodes/bifm_helper.jl")
...
 include("nodes/poisson.jl")
+include("nodes/or.jl")
+include("nodes/not.jl")
 
 # Deterministic nodes
 include("nodes/addition.jl")
 include("nodes/subtraction.jl")
 include("nodes/multiplication.jl")
+include("nodes/and.jl")
+include("nodes/implication.jl")
```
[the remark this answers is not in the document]
albertpod: Merge side effect, should be changed.

src/nodes/not.jl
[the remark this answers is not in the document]
albertpod: Same as for `&&`.

src/nodes/or.jl
```diff
@@ -0,0 +1,5 @@
+export OR
+
+struct OR end
+
+@node OR Deterministic [out, in1, in2]
```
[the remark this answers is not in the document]
albertpod: Should be `||` imho.

src/rules/and/marginals.jl
albertpod: Good point. I agree we should extend `contingency.jl`. Although in this case, this marginal is needed only for FE computations.

src/nodes/not.jl
bartvanerp: We use it in the multiplication node I believe

albertpod: True!

src/rules/and/marginals.jl
albertpod: Great suggestion!

ThijsvdLaar (commented):
Amazing effort, I agree with the comments by @bartvanerp Most importantly, the `Contingency` should indeed be normalized, since it represents a joint distribution over connected variables. It agree that it would be nice to write out the derivations in full, for checking, later reference, and most importantly, it can make for an impressive appendix in your thesis.

src/rules/and/marginals.jl
albertpod: Any ideas on how to introduce axes for this structure?

src/nodes/implication.jl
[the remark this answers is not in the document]
albertpod: @Chengfeng-Jia, what is your opinion on renaming the node?

src/rules/and/marginals.jl
bartvanerp: We could add these fields in the `Contingency` structure. However, I am not sure whether we can extract the variable names in the `rule` macro. I think we need to consult @bvdmitri for this.

src/nodes/implication.jl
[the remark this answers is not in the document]
Chengfeng-Jia: I think `IMPLY` may be a proper name. Besides, if we decide to use `&` instead of `AND`, can we use $\rightarrow$ to represent Implication.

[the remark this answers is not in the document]
Chengfeng-Jia: For the `Pointmass` part, thanks for @bartvanerp explain. I wonder if `Pointmass` can be regarded as a special case of Bernoulli distribution. For example, $x=1$ means $\mathcal{B}er\left(x \bigg| 1\right)$.

src/nodes/and.jl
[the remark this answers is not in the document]
bvdmitri: This is a nice feature, but requires an update to the `GraphPPL.jl`. Should not be too hard though, I think we can consider it as a new feature and do not include it in this pull request.

bvdmitri:
We need to add some sort of example to the documentation [Examples](https://biaslab.github.io/ReactiveMP.jl/stable/examples/overview/) section (and `demo/` folder). It not only shows how to use these new nodes, but also serves as an extra test because CI fails if some example from the documentation does not run properly.

albertpod:
@bartvanerp @bvdmitri @Chengfeng-Jia  I've added the demo demonstrating the usage of these nodes.

bvdmitri:
Thanks @albertpod , @Chengfeng-Jia  and @Sepideh-Adamiat . I'm fine with the current state of the PR. Once tests are passed we can merge.

bartvanerp (approved):

bartvanerp:
After adding docstrings with truth tables I think this PR should be ready for merge

bvdmitri (approved):
Great work @albertpod , @Chengfeng-Jia  and @Sepideh-Adamiat ! That is very good addition to our framework.

PR #175, opened 2022-07-29
fix preoptimised factorisation constraint rules for nodes with 2 inputs

bvdmitri:
This PR fixes and adds extra tests for preoptimised dispatch rules for unspecified constraints and a stochastic node with 2 inputs.

Details for triggering the bug:
- Node must be stochastic
- Node must have 1 input and 1 output
- Both interfaces must connect to random variables
- There must be no constraints specified (even empty)


albertpod (approved):
Funny one 👍

PR #194, opened 2022-08-18
Inverse Gamma Node

abpolym:
Adds the [inverse gamma](https://en.wikipedia.org/wiki/Inverse-gamma_distribution) node/distribution.


src/rules/gamma_inverse/out.jl
```diff
@@ -0,0 +1,9 @@
+export rule
+
+# Belief propagation
+# case: inbound messages on edges [α, θ] sum-product
+@rule GammaInverse(:out, Marginalisation) (m_α::Any, m_θ::Any) = GammaInverse(mean(m_α), mean(m_θ))
```
bvdmitri, line 5: This belief-propagation rule is correct only if both arguments are `PointMass` (remember that for belief propagation we generally do not use `Any`, if you spot somewhere most probably it is a mistake. For VMP in contrast `Any` is OK)
```suggestion
@rule GammaInverse(:out, Marginalisation) (m_α::PointMass, m_θ::PointMass) = GammaInverse(mean(m_α), mean(m_θ))
```

bvdmitri (changes requested):
The PR itself is very good, code is nice and clean. Well done! I spotted only one mistake (in the BP rule) and I would like for someone else to double-check all derivations. After that we can merge.

src/rules/gamma_inverse/marginals.jl
```diff
@@ -0,0 +1,13 @@
+export marginalrule
```
bvdmitri, line 1: minor but these exports aren't necessary, they are exported automatically (we should remove it from all other places)

albertpod:
Good job @abpolym! Once @bvdmitri comments are addressed, we can merge!

src/rules/gamma_inverse/marginals.jl
abpolym: I've removed both exports for rule and marginalrule

abpolym:
I've added the requested changes 😄

bvdmitri (approved):
Thanks @abpolym , well done.

PR #196, opened 2022-08-22
Fix mv-precision inference with known mean

### LINKED ISSUE #193: Wishart marginal not Hermitian in static model
Inference of a precision matrix in a static model produces a non-Hermitian Wishart marginal and a subsequent non-positive definite error when converting to PDMat object:
```
PosDefException: matrix is not positive definite; Cholesky factorization failed.
```

An example of the posterior scale matrix parameter before casting to PDMat is:
```
S = [-1.407 0.0599; -0.055 0.0023]
```
which is clearly not symmetric.

Example MWE:
```
# Sample size
N = 100

# True covariance matrix
Σ_true = [1.0 0.1;0.1 1.0]

# Generate data
X = randn(N,2)*cholesky(Σ_true).U

@model function modelΛ(N)
    
    X = datavar(Vector{Float64}, N)
    
    # Precision prior
    Λ ~ Wishart(3, diageye(2))
    
    # Likelihood
    for i in 1:N
        X[i] ~ MvNormalMeanPrecision(zeros(2), Λ)
    end
    
    return X, Λ
end

results = inference(
    model = Model(modelΛ, N),
    data = (X = [X[i,:] for i = 1:N],),
    returnvars = (Λ = KeepLast(),),
)

qΛ = results.posteriors[:Λ]
Σ_hat = inv(mean(qΛ))
```

We typically use the Wishart node to estimate process noise in dynamic model. In a static model, the problem does not occur at `N=1`, but starts to happen roughly 30% of the time at `N=10`, 50% at `N=20` and 100% at `N=100`. 

I suppose this is either a numerical instability that compounds over likelihood multiplications, or an error in the likelihood-based message from the`MvNormalMeanPrecision` node. Anyone else noticed this as well?

Versions:
- Julia: 1.7.2
- ReactiveMP: 2.4.1
- GraphPPL: 2.3.1

bvdmitri:
This PR fixes #193 by changing `WishartMessage` to store an inverse of its scale-matrix. This avoids unnecessary matrix invertions and correctly handles inference of precision matrix when mean is known. This fix potentially makes inference a bit faster too (way less inversions), but I haven't done any benchmarks. From the user perspective nothing changed really. The only one downside of this PR is that it is no longer possible to create an improper `Wishart` prior (because ReactiveMP.jl internally converts `Wishart` prior to `WishartMessage` which does one invertion).

### DIFF
--- a/src/distributions/wishart.jl
+++ b/src/distributions/wishart.jl
@@ -10,20 +10,23 @@ import SpecialFunctions: digamma
 
 Same as `Wishart` from `Distributions.jl`, but does not check input arguments and allows creating improper `Wishart` message.
 For model creation use `Wishart` from `Distributions.jl`. Regular user should never interact with `WishartMessage`.
+
+Note that internally `WishartMessage` stores (and creates with) inverse of its-scale matrix, but (for backward compatibility) `params()` function returns the scale matrix itself. 
+This is done for better stability in the message passing update rules.
 """
 struct WishartMessage{T <: Real, A <: AbstractMatrix{T}} <: ContinuousMatrixDistribution
-    ν::T
-    S::A
+    ν    :: T
+    invS :: A
 end
 
-function WishartMessage(ν::Real, S::AbstractMatrix{<:Real})
-    T = promote_type(typeof(ν), eltype(S))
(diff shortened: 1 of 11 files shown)


wmkouw (approved):
Nice work. I like that you added small static models to the test suite.

PR #201, opened 2022-08-24
Fix showprogress = true when iterations are not specified

### LINKED ISSUE #192: Error reporting for Progress in automatic inference API
In `inference(model=, ...)`, setting `showprogress=true` for a model that defaults to sum-product message passing will generate:
```
MethodError: no method matching Progress(::Nothing)
```
I first thought that I had forgotten to import ProgressMeter. I only realized later on that the model was silently defaulting to sum-product (it was previously using VMP and I made a small alteration to the model specification), in which, of course, no iteration occurs (and measuring progress doesn't make sense).

Generating a more descriptive error message might seem unnecessary, but I think it would help improve user experience. It's a small fix. I'll leave this open a few days in case someone wants to pick it up as their first issue. If not, I'll add it next week.

MWE:
```
@model function mwe()
    
    x = datavar(Float64)
    
    # Prior mean
    μ ~ NormalMeanVariance(0,1)
    
    # Likelihood
    x ~ NormalMeanVariance(μ,1)
    
    return x,μ
end

result = inference(
    model = Model(mwe),
    data = (x = 1.0,),
    returnvars = (μ = KeepLast(),),
    showprogress = true,
)
```

Versions:
- Julia: 1.7.2
- ReactiveMP: 2.4.1

bvdmitri:
This PR fixes #192

### DIFF
--- a/src/inference.jl
+++ b/src/inference.jl
@@ -116,10 +116,12 @@ end
 
 function Base.getproperty(result::InferenceResult, property::Symbol)
     if property === :free_energy && getfield(result, :free_energy) === nothing
-        error("""
-            Bethe Free Energy has not been computed. 
-            Use `free_energy = true` keyword argument for the `inference` function to compute Bethe Free Energy values.
-        """)
+        error(
+            """
+          Bethe Free Energy has not been computed. 
+          Use `free_energy = true` keyword argument for the `inference` function to compute Bethe Free Energy values.
+      """
+        )
     else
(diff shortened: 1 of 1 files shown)


bvdmitri:
@Sepideh-Adamiat thanks for the fix! Next time please, open a PR by yourself, otherwise I cannot approve your changes (GitHub does not allow me to approve my PR even though changes are from you)

PR #208, opened 2022-09-16
Add average energy for uniform dist

albertpod:
Adds average energy for the uniform node with corresponding tests.


Nimrais (approved):
I like it

bartvanerp:
@albertpod Let's double check this next week. You mentioned it would resemble the entropy of the distribution, but I am not 100% whether this is true. So we need to compute `- \int q(x) ln p(x) dx` (assuming `PointMass` `a` and `b`), but `q(x)` and `p(x)` do not have to be equal, as `a` and `b` are only imposed on `p(x)` and not on `q(x)`. Therefore it would better to represent a cross-entropy (which I couldn't find that quickly on Google).
Also, we need to specialize the average energy to `q_out::Uniform`.

albertpod:
@bartvanerp I am not sure what you mean.
Changed the interface for the score function.

bartvanerp (approved):

PR #209, opened 2022-09-19
add GPRegression by SSM demo

HoangMHNguyen:
add 2 notebooks on implementing GP regression by state-space model


bvdmitri:
@HoangMHNguyen , thanks for this demo! They are very quite good imo, but there are couple of points I would like you to address:
- Your branch was outdated with the current `master`, I merged `master` into your branch, don't forget to do `git pull` first when you make new changes
- Could you please merge two notebook into one? I would advise you to simply add `Matern 5/2` example right after the `Matern 3/2` as a new section in a single Jupyter notebook. Write something like: "Lets try to use a different kernel now..."
- You made an impressive work explaining the model specification and made useful references, however the demo does not have any result analysis. To me it seems a bit off, such a nice introduction and a simple output at the end. Could you please add a bit of words at the end as well. Not much, but smth like:
    - "We need `missing` rules because..."
    - "Here how we can plot out results..."
    - "The results make sense, because ..."

Overall looks very good, please address these minor comments and we will merge 👍

bvdmitri:
Documentation is failing due to the `PyPlot` library, which is unrelated to this PR.

Issue #172, opened 2022-07-25
Rule add-ons (such as scale factors)


bartvanerp:
With our recent acceptance to the SiPS conference on _Efficient Model Evidence Computation in Tree-structured Factor Graphs_, we would like to formalize the use of scale factors as prototyped in the `dev_scalefactors` branch. These scale factors basically specify the (log-)scaling of messages, which can be used for tracking the model evidence. Current implementations are limited to exact message passing procedures (sum product/belief propagation).

In the `dev_scalefactors` branch, scale factors have been implemented by creating messages `ScaledMessage`, which contain both the distribution as the scaling. The major downside of this approach is that it requires duplicating rules for this particular type of message. This adds a lot of clutter to the update rules and should be simplified.

After a discussion we propose the following rule specification:
```julia
@rule Node(messages, meta, options) = begin
    # compute parameters
    params = f(messages)
    @scaling begin
        # compute scale
        scale = g(messages, params)
        return scale
    end
    return new_message(params)
end
```
Inside the rule we can introduce a `@scaling` block, which specifies how the scaling gets updated in a factor node. This `@scaling` block will not be present in all rules, but just in a select few. Importantly, for efficiency purposes, this scaling block can use intermediate quantities of the message computation. The `options` argument can be used as a flag for the `@rule` macro to decide whether it wants to use the scaling update. It should only be used on command.

As discussed, it would be nice if this addition could be generalized not only to `@scaling`, but also to `@somethingelse`. This features should be enabled globally in the `@model` macro as `[ extras = Scaling() + Something() ]`. Whether we term this `extras`, `add-ons` or something else is up for debate.

This feature will also extend the `Message` (and `Marginal`) structs to something like
```julia
struct Message{D, E}
    distribution :: D
    extra :: E
end
```
where `extras` is `nothing` by default.

Preferable, the inference algorithm throws an error if a `@scaling` block is missing. Nodes (such as priors), with clamped interfaces, will always have a default scaling of 1 (or log-scaling of 0).

PR #215, opened 2022-09-26
Introducing addons for rules (+scale factors, +switch node, +mixture model)

bartvanerp:
For solving #172


src/rule.jl
```diff
@@ -292,13 +296,16 @@ macro rule(fform, lambda)
...
+                addonstype,
                 whereargs
             ) do
                 return quote
                     $(on_index_init)
                     $(m_init_block...)
                     $(q_init_block...)
-                    $(body)
+                    _addons = ()
+                    _message = $(MacroHelpers.remove_returns(body))
```
bvdmitri, line 307: Be careful with removing `return` statements from the whole message function body, as this may break things if a rule contains something like (even without addons)
```
@rule MyNode(...) (args...) = begin
    if condition
        return 1
    end
    return 2
end
```

bartvanerp: Resolved by [68809b4](https://github.com/biaslab/ReactiveMP.jl/pull/215/commits/68809b4f07b858321bd151dca5b7c3487c7c1ba2). It counts the number of return statements per macro by performing some string matching. Although it is better to be replaced with a proper `walk`, but I didn't manage to get this working.

bvdmitri: Yeah, that is better, I wonder if its really necessary to use `remove_returns` for the body of the `@rule` macro, but rather only on the body of the `@logscale` macro. I would probably move the `remove_returns` statement from `rule.jl:307` to `rule.jl:417` inside `macro logscale ... end`.

```suggestion
rule:jl:307 _message = body

rule.jl:417 return MacroHelpers.remove_returns(output)
```

I'm afraid that we did not really impose this restriction in the `@rule` macro in the past and it might be breaking. `@logscale` macro, however, is new and it may impose new restrictions on its body (such as no or only 1 return statement).

bvdmitri (approved):
@bartvanerp Great work!

PR #219, opened 2022-09-28
Develop delta fn eus fix indexed call rule

bvdmitri:
This PR enables test for indexed rules (+ includes all latest changes from the `master` branch).


bvdmitri:
CI failing is unrelated.

PR #221, opened 2022-09-29
Fix interaction between `autovar` and variables in a form of a tuple 

bvdmitri:
This PR fixes a bug when inference engine incorrectly attempted to create a node with an `AutoVar` object. Usually `AutoVar` object must be converted to either a new `RandomVariable` or reference an already existing variable. That didn't happen for the `tuple`-based variables, because of the incorrect type dispatch specification in the `make_node` function.


bartvanerp (approved):
Thanks @bvdmitri!

PR #235, opened 2022-11-04
224 gaussian mixture relax constraints all around

wouterwln:
Gaussian Mixture node rules are updated to leverage dispatch options and test cases are added.


bvdmitri (approved):
Well done @wouterwln !

PR #239, opened 2022-11-10
CVI stability fix

Nimrais:
This PR makes the CVI Approximation more stable and a bit more:

1. New optional parameter `proper_message`: ensuring that after the division, the message will be a proper distribution
2. adds more tests for the CVI rules
3. adds  a CVI parameter `grad`: the user can use it to change the differentiation procedure (e.g., compute hessian with Zygote)
4. fixes the lognormalizer/hessian sing issue
5. adds clarification comments
6. rename `render_cvi` -> `prod(approximation::ProdCVI, ...)`
7. the `ForwardDiff`hessian/grade computation use 10% less memory.
8. added a new hyperparameter for the CVI `n_gradpoints`: optional, defaults to 1, the number of points to estimate the likelihood gradient (dist*logp). It radically improves CVI precision (a gamma test to prove it added).

### DIFF
--- a/src/ReactiveMP.jl
+++ b/src/ReactiveMP.jl
@@ -178,12 +178,12 @@ function __init__()
 
         struct ZygoteGrad end
 
-        function compute_grad(::ZygoteGrad, A::F, vec_params) where {F}
-            Zygote.gradient(A, vec_params)[1]
+        function compute_grad(::ZygoteGrad, f::F, vec_params) where {F}
+            Zygote.gradient(f, vec_params)[1]
         end
 
-        function compute_hessian(::ZygoteGrad, A::G, ::F, vec_params) where {G, F}
-            Zygote.hessian(A, vec_params)
+        function compute_hessian(::ZygoteGrad, f::F, vec_params) where {F}
+            Zygote.hessian(f, vec_params)
         end
 
         get_df_m(::ZygoteGrad, ::Type{<:UnivariateNormalNaturalParameters}, ::Type{<:UnivariateGaussianDistributionsFamily}, logp_nc::Function) =
--- a/src/ReactiveMP.jl
+++ b/src/ReactiveMP.jl
@@ -196,7 +196,7 @@ function __init__()
         function compute_df_mv(::CVI{R, O, ForwardDiffGrad}, logp::F, vec::Vector) where {R, O, F}
             result = DiffResults.HessianResult(vec)
             result = ForwardDiff.hessian!(result, logp, vec)
-            return (DiffResults.gradient(result), DiffResults.hessian(result))
+            return DiffResults.gradient(result), DiffResults.hessian(result) ./ 2
         end
 
     end
--- a/src/ReactiveMP.jl
+++ b/src/ReactiveMP.jl
@@ -172,6 +172,20 @@ function __init__()
             return Flux.Optimise.update!(opt, vec(λ), vec(∇))
         end
     end
+
+    @require Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f" begin
+        export ZygoteGrad
+
+        struct ZygoteGrad end
+
+        function compute_grad(::ZygoteGrad, A::F, vec_params) where {F}
+            Zygote.gradient(A, vec_params)[1]
+        end
+
+        function compute_hessian(::ZygoteGrad, A::G, ::F, vec_params) where {G, F}
+            Zygote.hessian(A, vec_params)
+        end
+    end
 end
 
 end
--- a/src/ReactiveMP.jl
+++ b/src/ReactiveMP.jl
@@ -185,6 +185,12 @@ function __init__()
         function compute_hessian(::ZygoteGrad, A::G, ::F, vec_params) where {G, F}
             Zygote.hessian(A, vec_params)
         end
+
+        get_df_m(::ZygoteGrad, ::Type{<:UnivariateNormalNaturalParameters}, ::Type{<:UnivariateGaussianDistributionsFamily}, logp_nc::Function) =
+            (z) -> Zygote.gradient(logp_nc, z)[1]
+
+        get_df_v(::ZygoteGrad, ::Type{<:UnivariateNormalNaturalParameters}, ::Type{<:UnivariateGaussianDistributionsFamily}, logp_nc::Function, df_m::Function) =
+            (z) -> Zygote.hessian(logp_nc, z)
     end
 end
 
--- a/src/ReactiveMP.jl
+++ b/src/ReactiveMP.jl
@@ -178,19 +178,17 @@ function __init__()
 
         struct ZygoteGrad end
 
-        function compute_grad(::ZygoteGrad, f::F, vec_params) where {F}
+        function compute_gradient(::ZygoteGrad, f::F, vec_params::Vector) where {F}
             Zygote.gradient(f, vec_params)[1]
         end
 
-        function compute_hessian(::ZygoteGrad, f::F, vec_params) where {F}
+        function compute_hessian(::ZygoteGrad, f::F, vec_params::Vector) where {F}
             Zygote.hessian(f, vec_params)
         end
 
-        get_df_m(::ZygoteGrad, ::Type{<:UnivariateNormalNaturalParameters}, ::Type{<:UnivariateGaussianDistributionsFamily}, logp_nc::Function) =
-            (z) -> Zygote.gradient(logp_nc, z)[1]
-
-        get_df_v(::ZygoteGrad, ::Type{<:UnivariateNormalNaturalParameters}, ::Type{<:UnivariateGaussianDistributionsFamily}, logp_nc::Function, df_m::Function) =
-            (z) -> Zygote.hessian(logp_nc, z)
+        function compute_derivative(::ZygoteGrad, f::F, value::Real) where {F}
+            Zygote.gradient(f, value)[1]
+        end
     end
 end
 
--- a/src/ReactiveMP.jl
+++ b/src/ReactiveMP.jl
@@ -178,21 +178,13 @@ function __init__()
 
         struct ZygoteGrad end
 
-        function compute_gradient(::ZygoteGrad, f::F, vec_params::Vector) where {F}
-            Zygote.gradient(f, vec_params)[1]
-        end
-
-        function compute_hessian(::ZygoteGrad, f::F, vec_params::Vector) where {F}
-            Zygote.hessian(f, vec_params)
-        end
-
-        function compute_derivative(::ZygoteGrad, f::F, value::Real) where {F}
-            Zygote.gradient(f, value)[1]
-        end
+        compute_gradient(::ZygoteGrad, f::F, vec::AbstractVector) where {F} = Zygote.gradient(f, vec)[1]
+        compute_hessian(::ZygoteGrad, f::F, vec::AbstractVector) where {F}  = Zygote.hessian(f, vec)
+        compute_derivative(::ZygoteGrad, f::F, value::Real) where {F}       = Zygote.gradient(f, value)[1]
     end
 
     @require DiffResults = "163ba53b-c6d8-5494-b064-1a9d43ac40c5" begin
-        function compute_df_mv(::CVI{R, O, ForwardDiffGrad}, logp::F, vec::Vector) where {R, O, F}
+        function compute_df_mv(::CVI{R, O, ForwardDiffGrad}, logp::F, vec::AbstractVector) where {R, O, F}
             result = DiffResults.HessianResult(vec)
             result = ForwardDiff.hessian!(result, logp, vec)
             return DiffResults.gradient(result), DiffResults.hessian(result) ./ 2
--- a/src/ReactiveMP.jl
+++ b/src/ReactiveMP.jl
@@ -192,13 +192,11 @@ function __init__()
     end
 
     @require DiffResults = "163ba53b-c6d8-5494-b064-1a9d43ac40c5" begin
-        
         function compute_df_mv(::CVI{R, O, ForwardDiffGrad}, logp::F, vec::Vector) where {R, O, F}
             result = DiffResults.HessianResult(vec)
             result = ForwardDiff.hessian!(result, logp, vec)
             return DiffResults.gradient(result), DiffResults.hessian(result) ./ 2
         end
-
     end
 end
 
--- a/src/ReactiveMP.jl
+++ b/src/ReactiveMP.jl
@@ -190,6 +190,16 @@ function __init__()
             Zygote.gradient(f, value)[1]
         end
     end
+
+    @require DiffResults = "163ba53b-c6d8-5494-b064-1a9d43ac40c5" begin
+        
+        function compute_df_mv(::CVI{R, O, ForwardDiffGrad}, logp::F, vec::Vector) where {R, O, F}
+            result = DiffResults.HessianResult(vec)
+            result = ForwardDiff.hessian!(result, logp, vec)
+            return (DiffResults.gradient(result), DiffResults.hessian(result))
+        end
+
+    end
 end
 
 end
--- a/src/approximations/cvi.jl
+++ b/src/approximations/cvi.jl
@@ -36,14 +36,14 @@ Arguments
     Run `using Zygote` in your Julia session to enable the `ZygoteGrad` option support for the CVI `grad` parameter.
 
 """
-struct ProdCVI{R, O, G} <: AbstractApproximationMethod
+struct ProdCVI{R, O, G, B} <: AbstractApproximationMethod
     rng::R
     n_samples::Int
     num_iterations::Int
     opt::O
     grad::G
     warn::Bool
-    enforce_proper_messages::Val
+    enforce_proper_messages::Val{B}
 end
 
 get_grad(approximation::ProdCVI) = approximation.grad
--- a/src/approximations/cvi.jl
+++ b/src/approximations/cvi.jl
@@ -107,7 +107,7 @@ function render_cvi(approximation::CVIApproximation, logp_nc::F, initial) where 
         z_s = rand(rng, q_friendly)
 
         logq = (vec_params) -> logpdf(as_naturalparams(T, vec_params), z_s)
-        ∇logq = compute_grad(approximation.grad, logq, vec(λ))
+        ∇logq = compute_grad(get_grad(approximation), logq, vec(λ))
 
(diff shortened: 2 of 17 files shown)


src/distributions/normal.jl
```diff
@@ -479,7 +479,7 @@ function Base.:-(left::MultivariateNormalNaturalParameters, right::MultivariateN
 end
 
 function lognormalizer(η::UnivariateNormalNaturalParameters)
-    return η.weighted_mean^2 / (4 * η.minus_half_precision) + log(-2 * η.minus_half_precision) / 2
+    return -η.weighted_mean^2 / (4 * η.minus_half_precision) - log(-2 * η.minus_half_precision) / 2
 end
 
 function lognormalizer(η::MultivariateNormalNaturalParameters)
```
bartvanerp, line 485: perhaps using `chollogdet` in this function might be better performant (although this wasn't  part of the PR ;)

src/distributions/normal.jl
```diff
@@ -490,7 +490,7 @@ end
 # precludes the usage of logPdf functions previously defined. Below function is
 # meant to be used with Zygote.
 function Distributions.logpdf(η::UnivariateNormalNaturalParameters, x)
-    return log(invsqrt2π) + x * η.weighted_mean + x^2 * η.minus_half_precision + lognormalizer(η)
+    return log(invsqrt2π) + x * η.weighted_mean + x^2 * η.minus_half_precision - lognormalizer(η)
```
bartvanerp, line 493: Maybe replace `log(invsqrt2π)` with ` -log2π / 2` from `SpecialFunctions.jl`

src/distributions/normal.jl
```diff
@@ -538,13 +538,17 @@ end
...
+get_df_m(::ForwardDiffGrad, ::Type{<:MultivariateNormalNaturalParameters}, ::Type{<:MultivariateGaussianDistributionsFamily}, logp_nc::Function) =
+    (z) -> ForwardDiff.gradient(logp_nc, z)
 
-get_df_v(::Type{<:UnivariateNormalNaturalParameters}, ::Type{<:UnivariateGaussianDistributionsFamily}, df_m::Function) = (z) -> ForwardDiff.derivative(df_m, z)
+get_df_v(::ForwardDiffGrad, ::Type{<:UnivariateNormalNaturalParameters}, ::Type{<:UnivariateGaussianDistributionsFamily}, logp_nc::Function, df_m::Function) =
+    (z) -> ForwardDiff.derivative(df_m, z)
 
-get_df_v(::Type{<:MultivariateNormalNaturalParameters}, ::Type{<:MultivariateGaussianDistributionsFamily}, df_m::Function) = (z) -> ForwardDiff.jacobian(df_m, z)
+get_df_v(::ForwardDiffGrad, ::Type{<:MultivariateNormalNaturalParameters}, ::Type{<:MultivariateGaussianDistributionsFamily}, logp_nc::Function, df_m::Function) =
+    (z) -> ForwardDiff.jacobian(df_m, z)
```
bartvanerp, line 551: Can we get rid of these functions if we call `compute_grad()` or `compute_hessian` in the function body of `get_df_m` etc in `ReactiveMP.jl`? Let's discuss

src/ReactiveMP.jl
```diff
@@ -172,6 +172,26 @@ function __init__()
...
+
+        function compute_grad(::ZygoteGrad, A::F, vec_params) where {F}
+            Zygote.gradient(A, vec_params)[1]
+        end
+
+        function compute_hessian(::ZygoteGrad, A::G, ::F, vec_params) where {G, F}
+            Zygote.hessian(A, vec_params)
+        end
+
+        get_df_m(::ZygoteGrad, ::Type{<:UnivariateNormalNaturalParameters}, ::Type{<:UnivariateGaussianDistributionsFamily}, logp_nc::Function) =
```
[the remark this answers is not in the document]
bvdmitri: What is the purpose of `get_df_m`-like functions for Zygote? AFAIR these functions acted as a workaround specifically for the `ForwardDiff` library. If they are needed, there should be four functions, not two.

Also, as @bartvanerp mentioned already, the current implementation literally copy-pastes `compute_grad ` and `compute_hessian`. I think we discussed it already, but wouldn't it also be beneficial to compute gradient+hessian in one call (may be not a part of this PR though)?

src/approximations/cvi.jl
```diff
@@ -25,26 +25,37 @@ Arguments
  - `n_samples`: number of samples to use for statistics approximation
  - `num_iterations`: number of iteration for the natural parameters gradient optimization
  - `opt`: optimizer, which will be used to perform the natural parameters gradient optimization step
+ - `grad`: optional, structure to select how the gradient and the hessian will be computed
  - `warn`: optional, defaults to false, enables or disables warnings related to the optimization steps
+ - `proper_message`: optional, defaults to true, enables or disables inforce for the approximation be a proper distribution
```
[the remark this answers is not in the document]
bvdmitri: Agree with @bartvanerp , how about
```
- `enforce_proper_messages`: optional, defaults to true, ensures that a message, computed towards the inbound edges, is a proper distribution
```

src/approximations/cvi.jl
```diff
@@ -25,26 +25,37 @@ Arguments
  - `n_samples`: number of samples to use for statistics approximation
  - `num_iterations`: number of iteration for the natural parameters gradient optimization
  - `opt`: optimizer, which will be used to perform the natural parameters gradient optimization step
+ - `grad`: optional, structure to select how the gradient and the hessian will be computed
```
bvdmitri, line 28: Mention, that `ForwardDiff` is used by default

src/approximations/cvi.jl
```diff
@@ -25,26 +25,37 @@ Arguments
...
 """
-struct CVIApproximation{R, O} <: AbstractApproximationMethod
+struct CVIApproximation{R, O, G} <: AbstractApproximationMethod
     rng::R
     n_samples::Int
     num_iterations::Int
     opt::O
+    grad::G
     warn::Bool
+    proper_message::Bool
```
bvdmitri, line 46: I would make this field a `Val` based field, such that all checks could be compiled out:
https://docs.julialang.org/en/v1/manual/types/#%22Value-types%22

src/approximations/cvi.jl
```diff
@@ -58,6 +69,20 @@ const CVI = CVIApproximation
 # CVI implementations
 #---------------------------
 
+struct ForwardDiffGrad end
+
+function compute_grad(::ForwardDiffGrad, A::F, vec_params) where {F}
```
bvdmitri, line 74: How do these functions work in the combination with `get_df_m` and others?

bvdmitri (changes requested):
A very nice addition @Nimrais ! Please, address minor issues from me and Bart and we can merge.

src/ReactiveMP.jl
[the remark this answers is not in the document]
Nimrais: @bvdmitri I want to make our discussion clear for me, [the generic CVI step](https://github.com/biaslab/ReactiveMP.jl/blob/cvi-stability-fix/src/approximations/cvi.jl#L103-L119) is a 
```math
 \mathcal{F}^{-1}_{\lambda} \nabla_{\lambda} l
```
where $\mathcal{F}$ is the Fisher matrix, and $l$ is a likelihood. So we need the hessian of the [lognormalizer](https://github.com/biaslab/ReactiveMP.jl/blob/cvi-stability-fix/src/approximations/cvi.jl#L111) and gradient from the [likelihood](https://github.com/biaslab/ReactiveMP.jl/blob/cvi-stability-fix/src/approximations/cvi.jl#L109). I don't see how we can compute them at the same time.

So we can benefit from the computation of the hessian and the gradient only if the input message is a normal distribution. So you are proposing to create a function specifically for the normal case and call it kinda like `compute_df_dv`, am I correct?

src/approximations/cvi.jl
```diff
@@ -70,8 +95,8 @@ function render_cvi(approximation::CVIApproximation, logp_nc::F, initial) where
     hasupdated = false
 
     A = (vec_params) -> lognormalizer(as_naturalparams(T, vec_params))
-    gradA = (vec_params) -> ForwardDiff.gradient(A, vec_params)
-    Fisher = (vec_params) -> ForwardDiff.jacobian(gradA, vec_params)
+    gradA = (vec_params) -> compute_grad(approximation.grad, A, vec_params)
```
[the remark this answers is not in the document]
Nimrais: @bartvanerp Thanks, for a good suggestion. [Done.](https://github.com/biaslab/ReactiveMP.jl/blob/cvi-stability-fix/src/approximations/cvi.jl#L49)

src/approximations/cvi.jl
```diff
@@ -58,6 +69,20 @@ const CVI = CVIApproximation
...
+function compute_grad(::ForwardDiffGrad, A::F, vec_params) where {F}
+    ForwardDiff.gradient(A, vec_params)
+end
+
+function compute_hessian(::ForwardDiffGrad, A::G, ::F, vec_params) where {G, F}
+    ForwardDiff.hessian(A, vec_params)
+end
+
+function enforce_proper_message(inforce::Bool, λ::NaturalParameters, η::NaturalParameters)
+    return !inforce || (inforce && isproper(λ - η))
```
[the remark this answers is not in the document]
Nimrais: Yes, you are right, thanks!
[Done.](https://github.com/biaslab/ReactiveMP.jl/blob/cvi-stability-fix/src/approximations/cvi.jl#L85)

src/approximations/cvi.jl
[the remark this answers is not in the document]
bartvanerp: What I actually meant was `gradA = (vec_params) -> compute_grad(get_grad(approximation), A, vec_params)` :)

src/ReactiveMP.jl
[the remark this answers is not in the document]
bvdmitri: Yeah, in case if you take gradients of hessians of different things you cannot compute them together in one pass. I was assuming you compute it for one entity, but I was wrong.

bartvanerp:
FYI @bvdmitri today we discussed the PR and we mostly did some renaming. Most importantly, the `render_cvi(::CVIApproximation, ...)` function has been renamed to `prod(::ProdCVI, ...)`. We think this better covers the load of the method and should make it easier for outsiders to understand what is going on. Furthermore it should be easier to extend to more generic cases where we want to approximate some arbitrary product.

src/distributions/normal.jl
Nimrais: @bartvanerp Do you mean for the `MultivariateNormalNaturalParameters`?

src/distributions/normal.jl
Nimrais: We get rid of [these functions](https://github.com/biaslab/ReactiveMP.jl/blob/cvi-stability-fix/src/distributions/normal.jl#L541-L556). Right now, we are using only `compute_gradient`, `compute_hessian`, `compute_derivative`.

src/approximations/cvi.jl
Nimrais: We do not have the `get_df_m`-like functions now.

src/ReactiveMP.jl
```diff
@@ -172,6 +172,26 @@ function __init__()
...
+            (z) -> Zygote.gradient(logp_nc, z)[1]
```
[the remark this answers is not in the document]
Nimrais: Done

src/ReactiveMP.jl
```diff
@@ -172,6 +172,26 @@ function __init__()
...
+
+        get_df_v(::ZygoteGrad, ::Type{<:UnivariateNormalNaturalParameters}, ::Type{<:UnivariateGaussianDistributionsFamily}, logp_nc::Function, df_m::Function) =
+            (z) -> Zygote.hessian(logp_nc, z)
```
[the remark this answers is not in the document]
Nimrais: Done

src/approximations/cvi.jl
```diff
@@ -80,13 +105,14 @@ function render_cvi(approximation::CVIApproximation, logp_nc::F, initial) where
         z_s = rand(rng, q_friendly)
 
         logq = (vec_params) -> logpdf(as_naturalparams(T, vec_params), z_s)
-        ∇logq = ForwardDiff.gradient(logq, vec(λ))
+        ∇logq = compute_grad(approximation.grad, logq, vec(λ))
 
-        ∇f = Fisher(vec(λ)) \ (logp_nc(z_s) .* ∇logq)
+        fisher_matrix = Fisher(vec(λ))
+        ∇f = cholinv(fisher_matrix) * (logp_nc(z_s) .* ∇logq)
```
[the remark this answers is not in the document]
Nimrais: [Done](https://github.com/biaslab/ReactiveMP.jl/blob/cvi-stability-fix/src/approximations/cvi.jl#L141-L142)

src/approximations/cvi.jl
Nimrais: @bvdmitri Could you elaborate on this a bit? Do you mean we will create a structure:
```julia
structure ProperMessage{x} end

ProperMessage(x::bool) = ProperMessage{x}()

enforce_proper_messages(::ProperMessage{false}, ...) = false
enforce_proper_messages(::ProperMessage{true}, ...) = isproper(message)
```?

File: src/ReactiveMP.jl (hunk 4 above)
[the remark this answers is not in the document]
Comment (Nimrais) [reply src/ReactiveMP.jl]: @bvdmitri @bartvanerp I added a bit more effective hessian/grad computation for [the normal case](https://github.com/biaslab/ReactiveMP.jl/blob/cvi-stability-fix/src/ReactiveMP.jl#L196-L200).

Its working time is same. But it allocates 10% less.

You can reproduce my benchmark with this code:
```julia

using ReactiveMP
using BenchmarkTools
using ForwardDiff

struct EmptyOptimizer end

struct ForwardDiffOld end

function ReactiveMP.compute_gradient(::ForwardDiffOld, f::F, vec_params) where {F}
    ForwardDiff.gradient(f, vec_params)
end

function ReactiveMP.compute_hessian(::ForwardDiffOld, f::F, vec_params) where {F}
    ForwardDiff.hessian(f, vec_params)
end

meta = CVI(1, 100, EmptyOptimizer())
meta_old = CVI(1, 100, EmptyOptimizer(), ForwardDiffOld())

function sum_square(vec)
    return sum(map(x -> x^2, vec))
end

@benchmark ReactiveMP.compute_df_mv(meta_old, sum_square, $rand(20))
@benchmark ReactiveMP.compute_df_mv(meta, sum_square, $rand(20))
```

src/distributions/normal.jl
bartvanerp: Yess, if this `η.minus_half_precision` variable is a symmetric posdef matrix, then it is advantageous to first compute its Cholesky decomposition, which can then be used to both efficiently compute the `logdet` and the `\`.

src/distributions/normal.jl
```diff
@@ -538,37 +538,45 @@ end
...
+        df_μ1 = df_m - 2 * df_v * mean(q)
+        df_μ2 = df_v
         ∇f = as_naturalparams(T, df_μ1, df_μ2)
         ∇ = λ - η - ∇f
         λ_new = as_naturalparams(T, cvi_update!(opt, λ, ∇))
-        if isproper(λ_new)
+        if isproper(λ_new) && enforce_proper_message(approximation.enforce_proper_messages, λ_new, η)
             λ = λ_new
             hasupdated = true
         end
```
bartvanerp, line 582: Let's add some more comments here, similarly as in `cvi.jl`

src/approximations/cvi.jl
bvdmitri: There is a built-in `Val{true}/Val{false}` for that purpose.

src/distributions/normal.jl
```diff
@@ -538,37 +538,45 @@ end
...
+function compute_df_mv(approximation::CVI, logp_nc::F, z_s::Vector) where {F}
+    df_m = compute_gradient(get_grad(approximation), logp_nc, z_s)
+    df_v = compute_hessian(get_grad(approximation), logp_nc, z_s)
+    return df_m, df_v ./ 2
+end
 
-function render_cvi(approximation::CVIApproximation, logp_nc::F, initial::GaussianDistributionsFamily) where {F}
-    η = naturalparams(initial)
-    λ = naturalparams(initial)
+function prod(approximation::CVI, logp::F, dist::GaussianDistributionsFamily) where {F <: Function}
```
bvdmitri, line 558: I like this more and it also matches better with the already existing functionality. But 😅, the `prod` function in ReactiveMP currently expects a `pdf` (or a distribution), not `logpdf` as its argument. I'm making this comment for myself, no need to change anything @Nimrais , I'll deal with that discrepancy later myself during the final review.

src/approximations/cvi.jl
Nimrais: [Done](https://github.com/biaslab/ReactiveMP.jl/blob/cvi-stability-fix/src/approximations/cvi.jl#L28)

src/approximations/cvi.jl
Nimrais: [Done](https://github.com/biaslab/ReactiveMP.jl/pull/239/commits/7dfd256500c1c53446fb9d1c8e1722f43d2225e0)

src/distributions/normal.jl
Nimrais: [Done](https://github.com/biaslab/ReactiveMP.jl/pull/239/commits/2ed0b6e3cc56db2ea21cbe97e4945529e3b3613e)

src/distributions/normal.jl
Nimrais: Thanks for the nice catch! [Done](https://github.com/biaslab/ReactiveMP.jl/pull/239/commits/24c3808abbb5175dc073449dc9b26f9205f200d5)

bartvanerp (commented):
Nice work! PR is in good shape already. After passing the CI tests and after @bvdmitri's final inspection round, ready to merge in my opinion.

bvdmitri (approved):

bvdmitri:
Thanks @Nimrais (and @bartvanerp)! I'm running the tests locally, as soon as they pass I'll merge and tag a new release! Well Done!

bvdmitri:
Tests are failing

Nimrais:
> Tests are failing

Thanks! I missed that.

bvdmitri:
Great job @Nimrais! @bartvanerp thanks for helping & reviewing!

PR #257, opened 2023-02-03
Fix bernoulli

### LINKED ISSUE #255: Ambiguity with `probvec` for `Bernoulli` distribution
Not sure whether this is an issue, but I think it might be good to discuss this.

Consider `z = Categorical([0.1, 0.2, 0.7])`, which has classes `{1, 2, 3}`. If we call `probvec` on this distribution, we obtain `probvec(z) = [0.1, 0.2, 0.7]`, where the elements are nicely ordered according to the classes.

Now consider `y = Bernoulli(0.6)`, with classes `{0, 1}`. If we call `probvec` on this distribution, we obtain `probvec(y) = [0.6, 0.4]`. However, these correspond to the classes `1` and `0`, respectively. This ordering is different compared to the categorical one, although in this case we are ofc working with binary values.

This might cause some confusion and as a result some bugs might sneak in.

bartvanerp:
This PR fixes #255.

It has the following contributions:
- Reverses `probvec` of `Bernoulli` such that the ordering coincides with that of a `Categorical`.
- Fixes corresponding bug in `prod` function between `Bernoulli` and `Categorical`.
- Removes limitation of multiplying `Bernoulli` solely with `Categorical` of same size.
- Add/update tests.

### DIFF
--- a/src/distributions/bernoulli.jl
+++ b/src/distributions/bernoulli.jl
@@ -6,7 +6,7 @@ import StatsFuns: logistic
 
 vague(::Type{<:Bernoulli}) = Bernoulli(0.5)
 
-probvec(dist::Bernoulli) = (succprob(dist), failprob(dist))
+probvec(dist::Bernoulli) = (failprob(dist), succprob(dist))
 
 prod_analytical_rule(::Type{<:Bernoulli}, ::Type{<:Bernoulli}) = ProdAnalyticalRuleAvailable()
 
--- a/src/distributions/bernoulli.jl
+++ b/src/distributions/bernoulli.jl
@@ -23,8 +23,22 @@ end
 prod_analytical_rule(::Type{<:Bernoulli}, ::Type{<:Categorical}) = ProdAnalyticalRuleAvailable()
 
 function prod(::ProdAnalytical, left::Bernoulli, right::Categorical)
-    @assert length(probvec(right)) === 2 "Improper Bernoulli x Categorical product"
-    return prod(ProdPreserveType(Bernoulli), left, Bernoulli(first(probvec(right))))
+
+    # get probability vectors
+    p_left = probvec(left)
+    p_right = probvec(right)
+
+    # find length of new vector and compute entries
+    if length(p_left) >= length(p_right)
+        p_new = vcat(p_right..., zeros(length(p_left) - length(p_right)))
+        p_new .*= p_left
+    else
+        p_new = vcat(p_left..., zeros(length(p_right) - length(p_left)))
+        p_new .*= p_right
+    end
+
+    # return categorical with normalized probability vector
+    return Categorical(normalize!(p_new, 1))
 end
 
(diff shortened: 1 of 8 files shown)


src/distributions/bernoulli.jl
```diff
@@ -34,14 +48,27 @@ function prod(::AddonProdLogScale, new_dist::Bernoulli, left_dist::Bernoulli, ri
...
-    a = left_p * right_p + (one(left_p) - left_p) * (one(right_p) - right_p)
-    return log(a)
+function prod(::AddonProdLogScale, new_dist::Categorical, left_dist::Bernoulli, right_dist::Categorical)
+
+    # get probability vectors
+    p_left = probvec(left_dist)
+    p_right = probvec(right_dist)
+
+    # find length of new vector and compute entries
+    if length(p_left) >= length(p_right)
```
ismailsenoz, line 58: I think instead of if statement,  it  could be shorter to just write 
```
p_new = vcat(p_right..., zeros(abs(length(p_left) - length(p_right))))
Z = dot(p_left, p_new)

```

ismailsenoz (approved):
I only had one minor comment. But the pr is OK IMO.

src/distributions/bernoulli.jl
bartvanerp: nice suggestion, yet we either need to append zeros to `p_left` or `p_right` for which we can't get rid of the `if` statement

bvdmitri: @bartvanerp  I played around and this is what I came up with
```julia
function prod3(::ProdAnalytical, left::Bernoulli, right::Categorical)

    # get probability vectors
    p_left = probvec(left)
    p_right = probvec(right)

    # find the maximum length of both arguments
    mlength = max(length(p_left), length(p_right))
    
    # preallocate the result
    p_new = zeros(ReactiveMP.promote_samplefloattype(p_left, p_right), mlength)
    
    # Create 'e'xtended versions of the left and the right prob vectors
    # Do so by appending `0` with the Iterators.repeated, branch and allocation free
    e_left  = Iterators.flatten((p_left, Iterators.repeated(0, max(0, length(p_right) - length(p_left)))))
    e_right = Iterators.flatten((p_right, Iterators.repeated(0, max(0, length(p_left) - length(p_right)))))
    
    for (i, l, r) in zip(eachindex(p_new), e_left, e_right)
        @inbounds p_new[i] = l * r
    end

    # return categorical with normalized probability vector
    return Categorical(ReactiveMP.normalize!(p_new, 1))
end
```

```
@btime prod(ProdAnalytical(), $left, $right) # my version - 371.835 ns (5 allocations: 304 bytes)
@btime prod(ProdAnalytical(), $left, $right) # PR version - 487.253 ns (13 allocations: 496 bytes)
```

bartvanerp: Awesome!

bartvanerp:
Seems to have to do with comparing two `Categorical` distributions with `is_approx`. I have the feeling that this is an upstream issue, bc it only happens in Julia 1.6.

Issue #253, opened 2023-01-31
`MixtureModel` distribution from `Distributions.jl` not allowing `PointMass`


bartvanerp:
The `MixtureModel` distribution from `Distributions.jl` does not allow for having `PointMass` as one (or multiple) of its components. This is a result of `PointMass` not being a subtype of `Distribution`. I am not sure whether adding this supertype will break the internals in other parts of the inference function.

PR #260, opened 2023-02-03
Convert `MixtureModel` to `MixtureDistribution`

### LINKED ISSUE #256: Convert `MixtureModel` to `MixtureDistribution` carrying weights instead of a prior
As discovered by @ismailsenoz, the output distribution of the `Switch` node should not include a distribution on the prior attribute. Instead it should be a vector of values as the prior on these weights is marginalized out. Consecutive rules probably deal with this mistake in the correct way, so experiments/results will not affected.

bartvanerp:
This PR fixes #256 and includes the changes of PR #257.


bartvanerp:
Actually also resolves #253.

Issue #262, opened 2023-02-07
Rename `Switch` node to `Mixture`


bartvanerp:
After discussion in meeting, rename `Switch` node to `Mixture` node.

PR #263, opened 2023-02-07
Rename `Switch` to `Mixture`

bartvanerp:
Addresses #262.

Perhaps let's wait with the merge, because of recently submitted demo. Unless `Pluto` also tracks package versions ofc.


bvdmitri:
Pluto fixes versions and does not change them unless explicitly specified by a user so its fine. Perhaps we can also use the `@deprecated` macro

bvdmitri:
@bartvanerp I think the change is reasonable. Bump me when you want to merge this.

bvdmitri (approved):

PR #272, opened 2023-02-21
Additions to enable fix for RxInfer issue #9

wmkouw:
PR contains:
- Added `getpointmass` method for `Any` type.
- Added broadcastables for `MarginalSkipStrategy`'s
- Added `isproxy` and `isused` fields to `DataVariable` and `DataVariableCreationOptions` to distinguish between original datavars and those generated from mappings.
- Added tests on new fields.
- Small typo corrections.


wmkouw:
Do not merge. Still an issue with `getrecent(::ProxyObservable{..)`.

src/distributions/pointmass.jl
```diff
@@ -19,6 +19,7 @@ variate_form(::PointMass{M}) where {T, M <: AbstractMatrix{T}} = Matrixvariate
 sampletype(distribution::PointMass{T}) where {T} = T
 
 getpointmass(distribution::PointMass) = distribution.point
+getpointmass(point::Any) = point
```
bartvanerp, line 22: Might this cause issues when another distribution comes in? such as 
```julia
getpointmass(point::MvNormalMeanPrecision) = point
```
Would it instead be sufficient to have
```julia
getpointmass(::Real)
getpointmass(::AbstractArray)
```

bvdmitri: We discussed this with @wmkouw , but I think @bartvanerp has a valid concern. I didn't think of the `MvNormalMeanPrecision`. Perhaps it is better to constraint it to `Union{Real, AbstractArray}` indeed.

wmkouw: Cool. I will update it.

wmkouw:
Regarding `getrecent`: the newly constructed proxy datavar doesn't dispatch on `getrecent` in [`variables/data.jl:160`](https://github.com/biaslab/ReactiveMP.jl/blob/master/src/variables/data.jl#L160), because `typeof(as_marginal)` doesn't match the type of the anonymous function that `map` outputs.

Parametrizing the type of mapping,
```julia
function Rocket.getrecent(proxy::ProxyObservable{<:Marginal, S, M}) where {S, D, F, M <: Rocket.MapProxy{D, F}}
    return as_marginal(Rocket.getrecent(proxy.proxied_source))
end
```
fixes it (i.e., it successfully dispatches). But then `getrecent` doesn't apply the mapping; `Rocket.getrecent(proxy.proxied_source)` just returns the recent from `combineLatest`, i.e., a `Tuple{Marginal, Marginal}`, and `as_marginal` fails on a Tuple. 

I can implement `proxy.mappingFn(proxy.proxied_source)`, but you weren't a fan of this, @bvdmitri (because not all proxies carry `mappingFn` fields). What if I add a `getrecent` that dispatches on `MapProxy{<: Tuple, F}`?

bvdmitri:
@wouterwln 

This definition is not correct:
```
function Rocket.getrecent(proxy::ProxyObservable{<:Marginal, S, M}) where {S, D, F, M <: Rocket.MapProxy{D, F}}
    return as_marginal(Rocket.getrecent(proxy.proxied_source))
end
```

Dispatching on `MapProxy{<: Tuple, F}` will not solve the problem either. Leave the `getrecent` issue to me. I'll figure it out.

bvdmitri:
Linked branches: 
- RxInfer: https://github.com/biaslab/RxInfer.jl/tree/9-makenode-alldatavars
- Rocket: https://github.com/biaslab/Rocket.jl/tree/rx%239-makenode-alldatavars

bvdmitri:
@wmkouw  The issue with `getrecent` must have been solved in the `Rocket`, at least I could run tests locally. Please tell me if you see any other issues.

There is, however, one small issue that we need to address before we can merge. Currently tests shows a couple of warnings, like:

```
┌ Warning: Unused data variable has been found: 'a'. Ignore if 'a' has been used in deterministic nonlinear tranformation.
└ @ RxInfer ~/.julia/dev/RxInfer.jl/src/model.jl:172
┌ Warning: Unused data variable has been found: 'b'. Ignore if 'b' has been used in deterministic nonlinear tranformation.
└ @ RxInfer ~/.julia/dev/RxInfer.jl/src/model.jl:172
┌ Warning: Unused data variable has been found: 'a'. Ignore if 'a' has been used in deterministic nonlinear tranformation.
└ @ RxInfer ~/.julia/dev/RxInfer.jl/src/model.jl:172
┌ Warning: Unused data variable has been found: 'b'. Ignore if 'b' has been used in deterministic nonlinear tranformation.
└ @ RxInfer ~/.julia/dev/RxInfer.jl/src/model.jl:172
┌ Warning: Unused data variable has been found: '##nvar#325'. Ignore if '##nvar#325' has been used in deterministic nonlinear tranformation.
└ @ RxInfer ~/.julia/dev/RxInfer.jl/src/model.jl:172
┌ Warning: Unused data variable has been found: 'a'. Ignore if 'a' has been used in deterministic nonlinear tranformation.
└ @ RxInfer ~/.julia/dev/RxInfer.jl/src/model.jl:172
┌ Warning: Unused data variable has been found: 'b'. Ignore if 'b' has been used in deterministic nonlinear tranformation.
```

This happens because [here](https://github.com/biaslab/RxInfer.jl/blob/9-makenode-alldatavars/src/model.jl#L171) `RxInfer` thinks that if `datavars` are not connected, then they are unused and this was true. This PR changes that, datavars are not connected to any factor nodes, but are still used in the `combineLatest` operation for the proxy `datavar` which *is* connected to a factor node and is used in the factor graph.

We cannot simply do `nconnected += 1` as [here ](https://github.com/biaslab/ReactiveMP.jl/blob/6adf08ed0e7e687b2b155c712fe539a4ce887bb1/src/variables/data.jl#L167). `nconnected` field is used in the Free Energy computation and will produce wrong results if we discount number of connections. I think the easiest solution would be to add a new field, e.g. `used_implicitly` or something similar, which is `false` by default and becomes `true` if a datavar is used in the combination (datavar structure is mutable, so there is no problem to modify this field).

wmkouw:
Thank you very much, Dmitry. Quick work!

I have added an `isused` field to `DataVariable` and `DataVariableCreationOptions`, including tests. In RxInfer, I have suppressed the warning if a datavar is not connected but _is_ used. All tests pass on my end. I have created a draft PR at RxInfer as well.

bvdmitri (approved):

PR #277, opened 2023-02-22
Fix uniformscaling

### LINKED ISSUE #266: Creation and `prod` fails for `UniformScaling` and `Diagonal`
Some normal rules do not play nicely with `::LinearAlgebra.UniformScaling{Bool})` as a precision/covariance matrix.

bartvanerp:
This PR implements the following and closes #266 :
- Allows for using `UniformScaling` objects in `GraphPPL`.
- Implements multivariate normal constructors that convert `UniformScaling` objects into `Diagonal` matrices according to the length of the (weighted) mean vector.
- Further constrains the specialized `prod` function for multivariate normal distributions to exclude `Diagonal` matrices.

### DIFF
--- a/src/distributions/mv_normal_mean_covariance.jl
+++ b/src/distributions/mv_normal_mean_covariance.jl
@@ -26,7 +26,7 @@ function MvNormalMeanCovariance(μ::AbstractVector{T}) where {T}
     return MvNormalMeanCovariance(μ, convert(AbstractArray{T}, ones(length(μ))))
 end
 
-function MvNormalMeanCovariance(μ::AbstractVector{T1}, Σ::UniformScaling{T2}) where { T1, T2 }
+function MvNormalMeanCovariance(μ::AbstractVector{T1}, Σ::UniformScaling{T2}) where {T1, T2}
     T = promote_type(T1, T2)
     μ_new = convert(AbstractArray{T}, μ)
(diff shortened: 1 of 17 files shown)


src/distributions/mv_normal_mean_covariance.jl
```diff
@@ -88,7 +95,9 @@ function Base.prod(::ProdAnalytical, left::MvNormalMeanCovariance, right::MvNorm
     return MvNormalWeightedMeanPrecision(xi_left + xi_right, W_left + W_right)
 end
 
-function Base.prod(::ProdAnalytical, left::MvNormalMeanCovariance{T1}, right::MvNormalMeanCovariance{T2}) where {T1 <: LinearAlgebra.BlasFloat, T2 <: LinearAlgebra.BlasFloat}
+function Base.prod(
+    ::ProdAnalytical, left::MvNormalMeanCovariance{T1, <:AbstractVector, <:Matrix}, right::MvNormalMeanCovariance{T2, <:Vector, <:Matrix}
```
bvdmitri, line 99: Here is `AbstractVector`, but in the `MvNormalMeanPrecision` is `Vector`. Perhaps here should be `Vector` as well?

bartvanerp: You are absolutely right here. Should all be `AbstractVector`. I am fixing it right now.

bvdmitri (approved):

PR #282, opened 2023-02-27
Update rule Mixture and Transition

bartvanerp:
This PR implements:
- update Mixture rule according to paper
- add logscale for transition node

### DIFF
--- a/src/rules/transition/out.jl
+++ b/src/rules/transition/out.jl
@@ -10,6 +10,7 @@ end
 end
 
 @rule Transition(:out, Marginalisation) (m_in::DiscreteNonParametric, q_a::PointMass, meta::Any) = begin
+    @logscale 0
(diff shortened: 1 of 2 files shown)


src/rules/transition/out.jl
```diff
@@ -10,6 +10,7 @@ end
 end
 
 @rule Transition(:out, Marginalisation) (m_in::DiscreteNonParametric, q_a::PointMass, meta::Any) = begin
+    @logscale 0
```
bvdmitri, line 13: `@call_rule` should use the `getaddons` function. See for example: https://github.com/biaslab/ReactiveMP.jl/blob/e83dd3330292363ec7d356d8605ad77d13d24359/src/rules/multiplication/out.jl#L20

albertpod (approved):

PR #286, opened 2023-03-01
Add code for rules table

bartvanerp:
This PR adds code for printing all available rules, partly solving https://github.com/biaslab/RxInfer.jl/issues/74. Just not sure how to call the implemented `print_rules_table()` inside the docs of `RxInfer.jl`. I presume this table would deserve its own subsection under user guide - model specification at https://biaslab.github.io/RxInfer.jl/stable/manuals/model-specification/


bvdmitri:
@bartvanerp I've put your table to the documentation (in the `index.html` we should move it to other place later). One thing I noticed is that `meta` type is not present in the table. The table has "duplicate" rules which differ only in `meta` specification and `meta` by itself is quite important so I suggest to add it to the table as well.

bartvanerp:
I added the `meta` field and moved the location of the table with rules in the docs. Furthermore I added the table printing to  `src` and added `Markdown` into the dependencies. The main idea is that now everyone can just print the table in case they think they need it, but this is something that can be discussed.

src/rule.jl
```diff
@@ -1177,3 +1180,107 @@ function Base.showerror(io::IO, error::MarginalRuleMethodError)
...
+        txt *= inputs[k]
+        txt *= " | "
+        k == 1 ? txt *= meta : nothing
+        txt *= " |\n"
+    end
+    return txt
+end
+
+"""
+    Prints a table of all message passing update rules in ReactiveMP.
```
bvdmitri, line 1278:
```suggestion
    Prints a table of all message passing update rules in ReactiveMP.
    
    Use `Markdown.parse` on the output of this function to get a prettified table.
```

src/rule.jl
```diff
@@ -1177,3 +1180,107 @@ function Base.showerror(io::IO, error::MarginalRuleMethodError)
...
+"""
+function print_rules_table()
+    mtds = methods(ReactiveMP.rule)
+    Markdown.parse("""
```
bvdmitri, line 1282:
```suggestion
    """
```

src/rule.jl
```diff
@@ -1177,3 +1180,107 @@ function Base.showerror(io::IO, error::MarginalRuleMethodError)
...
+                   | Node | Output | Inputs | Meta |
+                   |:-----|:-------|:-------|:-----|
+                   """ * mapreduce(ReactiveMP.print_rule_rows, *, mtds))
```
bvdmitri, line 1285:
```suggestion
                   """ * mapreduce(ReactiveMP.print_rule_rows, *, mtds)
```

Project.toml
```diff
@@ -15,6 +15,7 @@ LazyArrays = "5078a376-72f3-5289-bfd5-ec5146d43c02"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 LoopVectorization = "bdcacae8-1622-11e9-2a5c-532679323890"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
+Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
```
bvdmitri, line 18:
```suggestion
```

src/rule.jl
```diff
@@ -26,6 +27,8 @@ This function is used to compute an outbound message for a given node
 - `addons`: Extra addons information
 - `__node`: Node reference
 
+For all available rules, call `ReactiveMP.print_rules_table()`.
```
bvdmitri, line 30:
```suggestion
For all available rules, see [`ReactiveMP.print_rules_table()`](@ref).
```

bvdmitri (commented):
I think we can ask user to add `Markdown` and call `Markdown.parse` by hand if they want prettified output, otherwise pure text (instead of adding `Markdown` as an explicit dependency).

PR #290, opened 2023-03-02
chore(): improve allocation profile for rule dispatching

### LINKED ISSUE #284: Excessive/unnecessary allocations in the `call_rule_make_node` and `MessageMapping`
The current implementation of the `rule` function accepts a node object as its last argument. None of the rules actually use the node object to compute a message, except for the delta node, which accesses the node object to get the node function. The functionality is nice, but these creates the following issues:

- `@call_rule` macro creates a node object even if the rule does not need it. It makes the procedure type-unstable. In a very simple example with rules for `+` node the execution time drops from `2ms` to `1ns` simply by removing the node argument.
- Only the `DeltaFn` rules actually use the node object, but the overhead in the `rule` exists for all other nodes. Simply by removing the node reference to `nothing` improves allocation profile and performance by a factor of 1.5-1.8 (timings are from the real-life private project).
 
There is no real reason to keep a reference to the node object for all message update rules, it can easily be fixed and we gain allocations/speed improvements almost for free.  I didn't test it for all possible examples we have, probably the improvement may be small for some examples, but this is worth doing anyway.

bvdmitri:
This PR fixes #284.

### DIFF
--- a/src/node.jl
+++ b/src/node.jl
@@ -866,7 +866,7 @@ function activate!(factornode::AbstractFactorNode, options)
             vmessageout = combineLatest((msgs_observable, marginals_observable), PushNew())  # TODO check PushEach
             vmessageout = apply_pipeline_stage(get_pipeline_stages(interface), factornode, vtag, vmessageout)
 
-            mapping = let messagemap = MessageMapping(fform, vtag, vconstraint, msgs_names, marginal_names, meta, addons, factornode)
+            mapping = let messagemap = MessageMapping(fform, vtag, vconstraint, msgs_names, marginal_names, meta, addons, node_if_required(fform, factornode))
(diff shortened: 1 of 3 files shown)


bvdmitri:
The allocation profile in the `TinyBenchmark.ipynb` from RxInfer:

`master` 
```
Memory estimate: 3.45 MiB, allocs estimate: 85342.
Memory estimate: 12.56 MiB, allocs estimate: 247416.
```

This PR:
```
Memory estimate: 3.03 MiB, allocs estimate: 83328.
Memory estimate: 11.10 MiB, allocs estimate: 241913.
```

albertpod:
I like the improvement. Looks elegant. I will wait for the tests locally and approve once it's green :)

albertpod (approved):
All tests are passing. Ready to merge.

Issue #280, opened 2023-02-23
Some tests for CVI do not converge


bvdmitri:
After #275 CI reports CVI warnings in test sets.

See for example: https://github.com/biaslab/ReactiveMP.jl/actions/runs/4243084222/jobs/7375403278

```
┌ Warning: CVI approximation has not updated the initial state. The method did not converge. Set `warn = false` to supress this warning.
└ @ ReactiveMP ~/work/ReactiveMP.jl/ReactiveMP.jl/src/approximations/cvi.jl:148
┌ Warning: CVI approximation has not updated the initial state. The method did not converge. Set `warn = false` to supress this warning.
└ @ ReactiveMP ~/work/ReactiveMP.jl/ReactiveMP.jl/src/distributions/normal.jl:624
┌ Warning: CVI approximation has not updated the initial state. The method did not converge. Set `warn = false` to supress this warning.
└ @ ReactiveMP ~/work/ReactiveMP.jl/ReactiveMP.jl/src/approximations/cvi.jl:148
┌ Warning: CVI approximation has not updated the initial state. The method did not converge. Set `warn = false` to supress this warning.
└ @ ReactiveMP ~/work/ReactiveMP.jl/ReactiveMP.jl/src/distributions/normal.jl:624
┌ Warning: CVI approximation has not updated the initial state. The method did not converge. Set `warn = false` to supress this warning.
└ @ ReactiveMP ~/work/ReactiveMP.jl/ReactiveMP.jl/src/approximations/cvi.jl:148
```

PR #291, opened 2023-03-03
Make CVI tests less confusing

Nimrais:
The following changes have been made in this pull request:

1. The convergence failed warning inside empty optimiser tests has been suppressed to avoid confusion.
2. The convergence warning has been included in all tests where we compare the CVI result with the analytical solution.
3. The tolerance level is smaller: we expect solutions to be more precise than it was before.

This PR resolves https://github.com/biaslab/ReactiveMP.jl/issues/280.


bvdmitri (approved):
Nice improvement @Nimrais !

PR #300, opened 2023-03-28
do not override Base.nameof

bvdmitri:
This PR removes the code that overrides the default behaviour of the `Base.nameof` as suggested in https://github.com/JuliaTesting/Aqua.jl/issues/95#issuecomment-1486526905 .

This makes it possible to test if `ReactiveMP` pirate some methods from other packages, which it actually does for methods like

```julia
mean(::typeof(log), distribution::Gamma) = ...
```

We need this methods and its technically the piracy, but should be safe. So I still keep `piracy = false` check from the `Aqua`.

### DIFF
--- a/src/marginal.jl
+++ b/src/marginal.jl
@@ -5,7 +5,7 @@ using Distributions
 using Rocket
 
 import Rocket: getrecent
-import Base: ==, ndims, precision, length, size, nameof, iterate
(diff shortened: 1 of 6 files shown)


albertpod (approved):
Looks good. Can be merged once CI is finished.

PR #302, opened 2023-03-30
Reimplement `@test_rules` and `@test_marginalrules`

bvdmitri:
This PR reimplements the `@test_rules` and `@test_marginalrules`. These both are quite old and were hacked around, especially the `with_float_conversions = true` option. This PR improves the code, adds tests and the documentation.

`with_float_conversions` has been renamed to `check_type_promotion`, technically breaking, but it is not a part of the public API, but rather our test infrastructure. 

The logic changed a little bit as well, thus I needed to tweak some rules to make the new code work. I'm actually quite confused why did it work before, but IMO new code is more robust. I also found one small bug! This PR also introduces a new family of promotion functions `promote_paramfloattype` in addition to `promote_samplefloattype`. Apparently in `Distributions.jl` the type of a sample is not necessarily aligns with the type of distribution parameters. 

Some un-used functions were removed.

Sorry for long list of changes, this is mainly due to renaming of the `with_float_conversions` option. The main files to check are: `distributions.jl`, `rule.jl` and their tests repsectively.


albertpod (approved):

albertpod:
@bvdmitri should we re-run CI?

bvdmitri:
The tests have been fixed and the PR is ready. Please wait for the CI and if everything is fine we can merge IMO.

PR #315, opened 2023-05-02
Fix GammaShapeRate pdf

Nimrais:
This PR resolves issue https://github.com/biaslab/ReactiveMP.jl/issues/314.


bvdmitri (approved):
Thanks!

PR #316, opened 2023-05-03
generalize bernoulli update rule and improve robustness

bartvanerp:
This PR does 2 things:
1. It generalizes the dispatching of the Bernoulli rules 
2. I believe it also increases robustness of the actual computations

With respect to 1, I am not sure whether this generalization is formally allowed, because there is no check for the domain of the variable over which we integrate. However, we currently also do not constrain the `PointMass` distribution.

### DIFF
--- a/src/rules/bernoulli/out.jl
+++ b/src/rules/bernoulli/out.jl
@@ -1,9 +1,13 @@
 export rule
 
-@rule Bernoulli(:out, Marginalisation) (m_p::Any,) = begin
+@rule Bernoulli(:out, Marginalisation) (m_p::Beta,) = begin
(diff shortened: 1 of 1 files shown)


bvdmitri:
As an idea we could implement "contracts" for the input arguments, e.g. check their domain etc. These contracts could be disabled (like asserts) for faster inference. We should keep this in mind for the next iteration of the `@rule` macro.

bvdmitri:
@bartvanerp I approve, but please add a comment regarding 1. above the computations. Just in case, so we know in the future that there possibly might be a limitation with the domain.

bvdmitri (approved):

PR #320, opened 2023-05-10
Add Extensions support from Julia 1.9

### LINKED ISSUE #303: The CVI re-use optimiser at each prod call, and this behaviour can be not suitable for some optimisers
Imagine you have an optimizer that has an inner state and somehow changes its parameters with each iteration (e.g., Adam).
If you will try to use it in several optimization tasks with a shared state: the convergence of each following task is not granted.

So if CVI is used in several consequent products with such an optimizer, it can start brake at some point.

The following code starts to find a bad approximation from some point:
```julia
using ReactiveMP
using StableRNGs
using Distributions
using Flux

n1 = NormalMeanVariance(10 * randn(rng), 10 * rand(rng))
n2 = NormalMeanVariance(10 * randn(rng), 10 * rand(rng))
n_analytical = prod(ProdAnalytical(), n1, n2)
cvi_result = prod(cvi, ContinuousUnivariateLogPdf((x) -> logpdf(n1, x)), n2)
@info n_analytical
@info prod(cvi, ContinuousUnivariateLogPdf((x) -> logpdf(n1, x)), n2)
@info isapprox(cvi_result, n_analytical, atol = 0.1)

for i in 1:10000
    cvi_result = prod(cvi, ContinuousUnivariateLogPdf((x) -> logpdf(n1, x)), n2)
    if !isapprox(cvi_result, n_analytical, atol = 0.1)
        @info "broken at $(i) iteration"
    end
end
```

This code works fine:
```julia
n1 = NormalMeanVariance(10 * randn(rng), 10 * rand(rng))
n2 = NormalMeanVariance(10 * randn(rng), 10 * rand(rng))
n_analytical = prod(ProdAnalytical(), n1, n2)
rng = StableRNG(42)
cvi = CVI(rng, 1, 1000, Flux.Adam(0.007), ForwardDiffGrad(), 10, Val(true), true)
cvi_result = prod(cvi, ContinuousUnivariateLogPdf((x) -> logpdf(n1, x)), n2)
@info n_analytical
@info prod(cvi, ContinuousUnivariateLogPdf((x) -> logpdf(n1, x)), n2)
@info isapprox(cvi_result, n_analytical, atol = 0.1)

for i in 1:10000
    cvi = CVI(rng, 1, 1000, Flux.Adam(0.007), ForwardDiffGrad(), 10, Val(true), true)
    cvi_result = prod(cvi, ContinuousUnivariateLogPdf((x) -> logpdf(n1, x)), n2)
    if !isapprox(cvi_result, n_analytical, atol = 0.1)
        @info "broken at $(i) iteration"
    end
end
```

Should we give a user an ability to reset an optimizer between iterations or prod calls?

bvdmitri:
This PR implements the [extensions](https://pkgdocs.julialang.org/dev/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)) feature instead of relying on the `Requires.jl` for Julia 1.9+. I also removed the `Flux` support with the `Optimisers.jl`. It means that all examples that previously used `Flux` for accessing optimizers now should use the `Optimisers.jl` instead. Technically might be breaking, but I don't really like that we depend on the whole `Flux` infrastructure for a simple optimisation method. In addition, it takes ages to compile `Flux` during the CI.

Partially fixes #303 as now it calls `deepcopy` on objects from `Optimisers.jl`, which solves half of the problem.

The next stage would be to do the same with `Optim`, `ForwardDiff`, `FastGaussQuadrature` and `HCubature`, but that would be another PR. For now I just experimented with packages that were part of the `Requires.jl` specification.


albertpod (approved):
Nice job @bvdmitri!

PR #321, opened 2023-05-11
add check for input arguments in the rules

### LINKED ISSUE #318: average energy does not check arguments and does not provide a useful error
Working on the transition mixture node, the following average energy function

```
@average_energy TransitionMixture (out_in_z::ContingencyTensor, B1::PointMass, B2::PointMass, B3::PointMass, B4::PointMass) = begin
 do magic
end
```
produces errors of this format which is not ideal
```
ERROR: MethodError: no method matching score(::AverageEnergy, ::Type{TransitionMixture}, ::
Val{(:out_in_z, :B1, :B2, :B3, :B4)}, ::Tuple{Marginal{ContingencyTensor{Float64, Array{Float64, 3}}, Nothing}, Marginal{PointMass{Matrix{Int64}}, Nothing}, Marginal{PointMass{Matrix{Int64}}, Nothing}, Marginal{PointMass{Matrix{Int64}}, Nothing}, Marginal{PointMass{Matrix{Int64}}, Nothing}}, ::Nothing)
```
The fix is simple (`B1 -> q_B1` and so on) but this is not apparent from the error. Issues opened at request of @bvdmitri to make sure we have records

bvdmitri:
This PR fixes #318

### DIFF
--- a/src/rule.jl
+++ b/src/rule.jl
@@ -398,7 +398,7 @@ macro rule(fform, lambda)
         return (iname, itype)
     end
 
-    rule_macro_check_fn_args(inputs; allowed_inputs = (:meta, ), allowed_prefixes = (:m_, :q_))
+    rule_macro_check_fn_args(inputs; allowed_inputs = (:meta,), allowed_prefixes = (:m_, :q_))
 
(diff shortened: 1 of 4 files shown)


albertpod (approved):
Approve, no concerns!

PR #322, opened 2023-05-22
add cholinv-family methods for UniformScaling

bvdmitri:
This PR adds `cholinv` and `cholsqrt` methods for `UniformScaling`. I noticed during some experiments, where some rules failed to take `cholinv`.

### DIFF
--- a/src/helpers/algebra/cholesky.jl
+++ b/src/helpers/algebra/cholesky.jl
@@ -5,9 +5,10 @@ using PositiveFactorizations
 
 import LinearAlgebra: BlasInt
 
-cholinv(x)           = inv(fastcholesky(x))
+cholinv(x) = inv(fastcholesky(x))
+cholinv(x::UniformScaling) = inv(x.λ) * I
 cholinv(x::Diagonal) = Diagonal(inv.(diag(x)))
-cholinv(x::Real)     = inv(x)
+cholinv(x::Real) = inv(x)
 
 function cholinv(x::AbstractMatrix{T}) where {T <: LinearAlgebra.BlasFloat}
     y = fastcholesky(x)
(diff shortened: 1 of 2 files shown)


src/helpers/algebra/cholesky.jl
```diff
@@ -43,9 +47,13 @@ function cholinv_logdet(x::AbstractMatrix{T}) where {T <: LinearAlgebra.BlasFloa
     # return inverse and log-determinant
     return y.factors, ly
 end
+
 cholinv_logdet(x::Diagonal) = Diagonal(inv.(diag(x))), mapreduce(z -> log(z), +, diag(x))
 cholinv_logdet(x::Real)     = inv(x), log(abs(x))
 
+fastcholesky(x::UniformScaling) = error("`fastcholesky` is not defined for `UniformScaling`. The shape is not determined.")
+fastcholesky!(x::UniformScaling) = error("`fastcholesky!` is not defined for `UniformScaling`. The shape is not determined.")
```
bartvanerp, line 55: Can't we do `fastcholesky(x::UniformScaling) = isposdef(x) ? Cholesky(sqrt(x.λ) * I, 'L', convert(BlasInt, 0)) : Error()`?

bartvanerp (commented):
Great work!

src/helpers/algebra/cholesky.jl
bvdmitri: Unfortunately no, because `factors` in the `Cholesky` structure must be of type `AbstractMatrix`, but `sqrt(x.λ) * I` is not an `AbstractMatrix`

src/helpers/algebra/cholesky.jl
```diff
@@ -5,23 +5,26 @@ using PositiveFactorizations
...
-cholsqrt(x)           = Matrix(fastcholesky(x).L)
+cholsqrt(x) = Matrix(fastcholesky(x).L)
+cholsqrt(x::UniformScaling) = sqrt(x.λ) * I
 cholsqrt(x::Diagonal) = Diagonal(sqrt.(diag(x)))
-cholsqrt(x::Real)     = sqrt(x)
+cholsqrt(x::Real) = sqrt(x)
 
-chollogdet(x)           = logdet(fastcholesky(x))
+chollogdet(x) = logdet(fastcholesky(x))
+chollogdet(x::UniformScaling) = error("logdet is not defined for `UniformScaling`")
```
bartvanerp, line 25: `det` is defined (although only for two cases, see https://github.com/JuliaLang/julia/blob/f44be79d57c6e8c07e6abb58c39ebbbe65ad4a0e/stdlib/LinearAlgebra/src/uniformscaling.jl#L239-L247) and for `logdet` it then uses https://github.com/JuliaLang/julia/blob/f44be79d57c6e8c07e6abb58c39ebbbe65ad4a0e/stdlib/LinearAlgebra/src/generic.jl#L1701. So I propose to change this to `chollogdet(x::UniformScaling) = logdet(x)`

bartvanerp (approved):

PR #327, opened 2023-05-26
Make the sample list approximation more generic

### LINKED ISSUE #323: approximate_as_samplelist procedure uses `eltype` instead of `paramfloattype`
This causes problems with container initialisation in the
https://github.com/biaslab/ReactiveMP.jl/blob/master/src/distributions/sample_list.jl#L260

Encountered by @ThijsvdLaar

bvdmitri:
This PR fixes #323 and adds non-included tests for the `function.jl` (which were failing btw, oops).


albertpod (approved):
good catch @bvdmitri

PR #338, opened 2023-07-06
Add Half-normal node

albertpod:


albertpod:
ping @bvdmitri

albertpod:
@bvdmitri note that there are no tests for prod of HalfNormal and Gamma, besides, I can't figure out why `check_type_promotion = true` fails.

I've also fixed the namings in `InverseGamma` tests.

bvdmitri:
The fail in `check_type_promotion` is probably due to the `Inf` which is of type `Float64`

albertpod:
ping @bvdmitri

bvdmitri (approved):

PR #354, opened 2023-10-09
Switch to FastCholesky.jl

bvdmitri:
This PR removes the `cholesky` folder and uses FastCholesky instead


bartvanerp (approved):
Simplified `distributions/wishart_inverse.jl` slightly. If tests pass, feel free to merge.

PR #360, opened 2023-10-27
Upate for PointMass initmarginals and initmasseges

Chengfeng-Jia:
For the RxInfer [issue](https://github.com/biaslab/RxInfer.jl/issues/101).


bartvanerp (commented):
Good that we identified the related lines, let's write two small tests and get it merged

bvdmitri (commented):
Ah, that is the reason, good that you've found it. A small test would suffice.

Chengfeng-Jia:
Thank @bvdmitri for your previous extensible code, which makes writing a test easier.

bartvanerp (approved):

PR #365, opened 2023-11-24
Allow const/data inputs for the non-linear deterministic nodes

bvdmitri:
This PR (potentially) fixes https://github.com/biaslab/RxInfer.jl/issues/30


bvdmitri:
Linked PR https://github.com/biaslab/RxInfer.jl/pull/193

bvdmitri:
The tests will fail until the new package is registered in the global Julia registry https://github.com/JuliaRegistries/General/pull/95797

bartvanerp (approved):
I am a bit fuzzy on the details, but the accompanying tests are convincing enough. Great work!

bartvanerp:
One quick question before merge, does this also work when values are directly inserted in the function definition, e.g. `y ~ g(x, 3.0)`? Perhaps it already did, but just checking

bvdmitri:
> One quick question before merge, does this also work when values are directly inserted in the function definition, e.g. `y ~ g(x, 3.0)`? Perhaps it already did, but just checking

It does work, but the input to the node must be a constvar object. GraphPPL.jl does it for you automatically and converts `3.0` to `constvar(:blabla, 3.0)`.

PR #367, opened 2023-12-07
Fix unscented transform for univariate input multivariate output

ismailsenoz:


albertpod:
It appears that this pull request is currently out of sync with the main branch. Additionally, I don't see any tests associated with the changes made. Could you clarify what the issue was previously?

ismailsenoz:
Unscented transform was not working for univariate input multivariate output cases. It is most likely out of sync. You should take a look at the dev branches of expectation approximations for a better implementation of unscented transforms. I will close this pr.

PR #370, opened 2024-01-05
Add rules and tests for MvNormalWeightedMeanPrecision node

albertpod:
This PR encompasses the following contributions for the `MvNormalWeightedMeanPrecision` node:

1. Fix for average energy.
2. Addition of rules for the out interface (particularly useful for priors).
3. Inclusion of missing tests.

### DIFF
--- a/src/nodes/mv_normal_weighted_mean_precision.jl
+++ b/src/nodes/mv_normal_weighted_mean_precision.jl
@@ -5,5 +5,5 @@ import StatsFuns: log2π
 @average_energy MvNormalWeightedMeanPrecision (q_out::Any, q_ξ::Any, q_Λ::Any) = begin
     m_ξ, v_ξ     = mean_cov(q_ξ)
     m_out, v_out = mean_cov(q_out)
-    return (ndims(q_out) * log2π - mean(chollogdet, q_Λ) + tr(mean(q_Λ) * (m_out * m_out' + v_out)) - 2*dot(m_out,m_ξ) + tr(mean(cholinv, q_Λ) * (m_ξ * m_ξ' + v_ξ))) / 2
+    return (ndims(q_out) * log2π - mean(chollogdet, q_Λ) + tr(mean(q_Λ) * (m_out * m_out' + v_out)) - 2 * dot(m_out, m_ξ) + tr(mean(cholinv, q_Λ) * (m_ξ * m_ξ' + v_ξ))) / 2
 end
--- a/src/nodes/mv_normal_weighted_mean_precision.jl
+++ b/src/nodes/mv_normal_weighted_mean_precision.jl
@@ -3,17 +3,11 @@ import StatsFuns: log2π
 @node MvNormalWeightedMeanPrecision Stochastic [out, (ξ, aliases = [xi, weightedmean]), (Λ, aliases = [invcov, precision])]
 
 @average_energy MvNormalWeightedMeanPrecision (q_out::Any, q_ξ::Any, q_Λ::Any) = begin
-    m_mean, v_mean = mean_cov(q_ξ)
-    m_out, v_out   = mean_cov(q_out)
-    return @views (ndims(q_out) * log2π - mean(logdet, q_Λ) + tr(mean(q_Λ) * (v_out + v_mean + (m_out - m_mean) * (m_out - m_mean)'))) / 2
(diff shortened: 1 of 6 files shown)


src/nodes/mv_normal_weighted_mean_precision.jl
```diff
@@ -5,15 +5,15 @@ import StatsFuns: log2π
 @average_energy MvNormalWeightedMeanPrecision (q_out::Any, q_ξ::Any, q_Λ::Any) = begin
     m_mean, v_mean = mean_cov(q_ξ)
     m_out, v_out   = mean_cov(q_out)
-    return (ndims(q_out) * log2π - chollogdet(mean(q_Λ)) + tr(mean(q_Λ) * (v_out + v_mean + (m_out - m_mean) * (m_out - m_mean)'))) / 2
```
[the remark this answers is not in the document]
bartvanerp: I have some concerns as to whether this rule is actually correct (likely my bad). Just did the derivations again and seem to end up with 
`ndims(q_out) * log2π - mean(chollogdet, q_Λ) + tr(mean(q_Λ) * (v_out + v_mean + m_mean*m_mean'+ m_out * m_out')) - 2*m_out*m_mean ) / 2`
I might have a mistake somewhere, but we better double check.

src/nodes/mv_normal_weighted_mean_precision.jl
```diff
@@ -5,15 +5,15 @@ import StatsFuns: log2π
...
+    return @views (ndims(q_out) * log2π - mean(logdet, q_Λ) + tr(mean(q_Λ) * (v_out + v_mean + (m_out - m_mean) * (m_out - m_mean)'))) / 2
```
bartvanerp, line 8: Good catch! Would it make sense to add the method `mean(chollogdet, ...)` to `BayesBase.jl`? Then we can use the symmetry information to speed up this computation.

bartvanerp, line 8: Why the @views? Seems unnecessary to me

bartvanerp, line 8: I have some concerns as to whether this rule is actually correct (likely my bad). Just did the derivations again and seem to end up with
ndims(q_out) * log2π - mean(chollogdet, q_Λ) + tr(mean(q_Λ) * (v_out + v_mean + m_mean*m_mean'+ m_out * m_out')) - 2*m_out*m_mean ) / 2
I might have a mistake somewhere, but we better double check.

albertpod: I've changed it to a different form.

bvdmitri:
Something is completely wrong with this PR, I disagree with the change as for now. I also think that the tests might be incorrect. Lets discuss. But tldr, we cannot simply call average energy of the mean-precision parametrization, because semantics of the inputs is different.

albertpod:
Thanks, @bvdmitri and @bartvanerp, for spotting the mistakes. It's a little more difficult than I thought.

Anyway, just to re-iterate, I want to use this node as a prior mostly.

src/nodes/mv_normal_weighted_mean_precision.jl
```diff
@@ -3,17 +3,7 @@ import StatsFuns: log2π
...
-    m, V = mean_cov(q_out_ξ)
-    d = div(ndims(q_out_ξ), 2)
-    return @views (
-        d * log2π +
-        -chollogdet(mean(q_Λ)) +
-        tr(mean(q_Λ) * (V[1:d, 1:d] - V[1:d, (d + 1):end] - V[(d + 1):end, 1:d] + V[(d + 1):end, (d + 1):end] + (m[1:d] - m[(d + 1):end]) * (m[1:d] - m[(d + 1):end])'))
-    ) / 2
+    m_ξ, v_ξ     = mean_cov(q_ξ)
+    m_out, v_out = mean_cov(q_out)
+    return (ndims(q_out) * log2π - mean(logdet, q_Λ) + tr(mean(q_Λ) * (m_out * m_out' + v_out)) - 2m_out'm_ξ + tr(mean(inv, q_Λ) * (m_ξ * m_ξ' + v_ξ))) / 2
```
bartvanerp, line 8:
```suggestion
    return (ndims(q_out) * log2π - mean(chollogdet, q_Λ) + tr(mean(q_Λ) * (m_out * m_out' + v_out)) - 2*dot(m_out,m_ξ) + tr(mean(cholinv, q_Λ) * (m_ξ * m_ξ' + v_ξ))) / 2
```

PR #371, opened 2024-01-08
Update LICENSE

bartvanerp:


bvdmitri (commented):
Maybe we simply keep the original year, like 2021. I googled it a bit and it seems its fine to have only the initial year. Otherwise we can also setup this https://github.com/marketplace/actions/update-license-copyright-year-s

albertpod:
I'm fine with start-present

bvdmitri (approved):

PR #372, opened 2024-01-09
Use Julia 1.10 for tests

bvdmitri:
This PR changes CI to use Julia 1.10 for tests


bvdmitri:
@albertpod There is an issue with the continuous transition node, not sure what is causing it.

PR #375, opened 2024-01-10
Change rules for the uninformative node

bvdmitri:
Another attempt to #373 . Spotted by @wouterwln & @bartvanerp . New logic for the predictions did not take into account that someone would want to define a rule that accepts `Missing` as an argument, which is sent out of `Uninformative` node.  This PR changes the outbound message rule for the `Uninformative` node and allows users to define their rules with `Uninformative` arguments.


bvdmitri:
This PR passes all the tests in RxInfer

albertpod:
Nice! Please merge once the tests pass.

albertpod (approved):

PR #403, opened 2024-06-17
Allow functional form constraints that are not sub-typed from AbstractFormConstraint

bvdmitri:
This PR enables functional form constraints that do not subtype `AbstractFormConstraint`. This is useful for supporting objects defined in different packages, such as `ProjectedTo` from `ExponentialFamilyProjection.jl`.


albertpod (approved):
LGTM!

Issue #402, opened 2024-06-17
SoftDot node does not have structured rules


albertpod:
The current `SoftDot` node implementation only permits mean-field factorization between states, as shown below:
```julia
constraints = @constraints begin
  q(x, y, θ, γ) = q(x)q(y)q(θ)q(γ)
end;
```
This is likely because it was designed primarily for likelihood cases. However, including structured rules could be beneficial. This would require implementation and allow for the following factorization:
```julia
constraints = @constraints begin
  q(x, y, θ, γ) = q(x, y)q(θ)q(γ)
end;
```
Interestingly, `ReactiveMP.jl` already has this factorization and corresponding rules (just hidden :) ). `SoftDot` is essentially a univariate output of `AR` node. Since `AR` nodes (designed for state space models) copy previous states (buffering), instead of reimplementing all new structured `SoftDot` rules, one could use the `@call_rule` macro to call AR node rules and select only the first output element.

bvdmitri:
Fixed in #404

PR #404, opened 2024-06-25
Add structured VMP rules for softdot node

albertpod:
This PR addresses issue #402 by introducing new set of rules for the `SoftDot` node. The `SoftDot` node is a variant of the autoregressive node but operates without buffering. Therefore, the rules and tests for `SoftDot` are the same as for other messages, except for the backward message to the `x` interface. Average energy is identical to AR node, but does not require entropy correction anymore.

Additionally, I have added error handling to the `dot` product rules to suggest using `SoftDot` when a closed-form solution is not available.


bvdmitri (approved):
Thanks @albertpod !

Issue #401, opened 2024-05-31
ContinuousTransition node does not have mean-field rules


albertpod:
The current implementation of the continuous transition node allows only structured factorization between the states, i.e.,
```julia
constraints = @constraints begin
    q(x, y, h, Λ) = q(x, y)q(h)q(Λ)
end;
```

While this factorization is beneficial for state space models (as it introduces coupling between states), it doesn't allow one of $x$ or $y$ to be observed (clamped). The implementation of these (MF) rules needs to be added, which will also allow the following factorization:
```julia
constraints = @constraints begin
    q(x, y, h, Λ) = q(x)q(y)q(h)q(Λ)
end;
```

bvdmitri:
Fixed in #406

PR #406, opened 2024-06-27
Add MF rules for CTransition

albertpod:
This PR addresses issue #401  by introducing a new set of rules for the `CTransition`node. Mainly, the new rules enable mean-field constraint around the node.

### DIFF
--- a/src/nodes/predefined/continuous_transition.jl
+++ b/src/nodes/predefined/continuous_transition.jl
@@ -122,7 +122,7 @@ end
     return AE
 end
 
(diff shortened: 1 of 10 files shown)


bvdmitri:
Can we add docstrings to the ContinuousTransition and some caveats (see https://github.com/ReactiveBayes/ReactiveMP.jl/issues/400)?  I guess one of them would be the fact that this node requires marginal on `a` to compute the message on `a`.

albertpod:
Thanks @bvdmitri! Will do.

bvdmitri (approved):
Great, thanks @albertpod !

PR #410, opened 2024-07-22
Fix poisson average energy computation

ismailsenoz:
Average energy computation for Poisson node assumed samples are only positive. However, Poisson samples are non-negative. This PR fixes that by starting the average energy computation from 0.


bvdmitri (approved):
Good catch!

PR #412, opened 2024-08-07
remove msg_dep and mar_dep functions

wouterwln:
After a discussion with @ThijsvdLaar , we discovered that those 2 functions are only ever used inside themselves, and are not used in any `RxInfer` or `ReactiveMP` tests. I'm not sure if these functions are necessary or if they can be deleted, but it seems that `functional_dependencies` does what these functions were supposed to do.

### DIFF
--- a/src/nodes/dependencies.jl
+++ b/src/nodes/dependencies.jl
@@ -111,30 +111,6 @@ end
 
 RequireMessageFunctionalDependencies(; kwargs...) = RequireMessageFunctionalDependencies((; kwargs...))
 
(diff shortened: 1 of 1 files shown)


bvdmitri:
Ah, you might be right. Can you double check `RxInfer` tests with the `make devtest` and `make devexamples` [commands](https://reactivebayes.github.io/RxInfer.jl/stable/contributing/guidelines/#Makefile)?

wouterwln:
Both work!

bvdmitri (approved):

PR #413, opened 2024-08-16
Adapt out message rule for delta node to new syntax of ExponentialFamilyProjection

ismailsenoz:
This PR changes the out message rule for delta node such that the rule uses `project_to` from `ExponentialFamilyProjection` and simplifies the code.


bvdmitri (approved):
woow the code looks super nice now, thanks @ismailsenoz

bvdmitri:
tests in RxInfer are failing, I'll look into it

PR #418, opened 2024-10-02
Relax rules for structured VMP in Transition node

wouterwln:
With `UnfactorizedData`, we should be able to send `PointMass` messages with a factorized `a` as well. The rules are the same as for a Categorical (as far as I can see) but we didn't dispatch on them yet.

### DIFF
--- a/src/rules/transition/in.jl
+++ b/src/rules/transition/in.jl
@@ -1,6 +1,6 @@
 import Base.Broadcast: BroadcastFunction
 
-@rule Transition(:in, Marginalisation) (m_out::Categorical, m_a::PointMass) = begin
+@rule Transition(:in, Marginalisation) (m_out::Union{DiscreteNonParametric, PointMass}, m_a::PointMass) = begin
     @logscale log(sum(mean(A)' * probvec(m_out)))
     p = mean(m_a)' * probvec(m_out)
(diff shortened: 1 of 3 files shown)


ThijsvdLaar:
Does this change rely on an implementation of `BayesBase.probvec(::PointMass)`? Since the point mass rules are only supposed to work with one-hot encoded vectors, how is this handled?

bvdmitri:
I think the current `probvec` simply returns the underlying vector without checking if its one hot encoded or not:
```julia
julia> using BayesBase

julia> probvec(PointMass(1))
ERROR: probvec(::PointMass{ <: Real }) is not defined
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] probvec(::PointMass{Int64})
   @ BayesBase ~/.julia/dev/BayesBase/src/densities/pointmass.jl:68
 [3] top-level scope
   @ REPL[2]:1

julia> probvec(PointMass([1, 2]))
2-element Vector{Int64}:
 1
 2
```

ThijsvdLaar:
Ok, seems reasonable to not check for one-hot encoding, maybe make a note of this for clarity.

ThijsvdLaar (approved):

PR #419, opened 2024-10-02
Fix typo in DeferredMessage

wouterwln:
Was spelled DefferedMessage


bvdmitri (commented):
Thanks Wouter! I guess also need to be changed in the documentation? CI should probably fail if the docs document a non-existing symbol

bvdmitri (approved):
Nice!

PR #428, opened 2024-11-06
CVIProjection with constraints on interfaces

Nimrais:
# Added flexibility to projection forms in CVI rules

### Enhanced CVIProjection with projection forms control
- Added `out_prjparams` parameter to control output edge projection form
- Added `in_prjparams` parameter as a NamedTuple to control input edge projection forms

### Added flexible form selection mechanism
- Created `get_kth_in_form` to safely access input forms by index
- Forms are optional - if not specified, the system uses the default form inferred from the marginal
- Added smart form matching to maintain optimal performance:
  - Uses efficient supplementary distribution when forms match or default form is used
  - Falls back to including logpdf in objective only when forms truly differ

### Optimized projection logic
- For output edge:
  - Uses specified form if provided
  - Falls back to inferred form from marginal if not specified
- For input edges:
  - Allows per-edge form specification through NamedTuple
  - Maintains optimal performance by detecting form compatibility
  - Handles form mismatches gracefully by incorporating logpdf in objective

## Example Usage

```julia
# Specify output form
form_out = ProjectedTo(Normal)

# Specify forms for specific input edges
forms_in = (in_2 = ProjectedTo(MvNormalMeanCovariance, 2),)

# Create CVIProjection with custom forms
meta = DeltaMeta(
    method = CVIProjection(
        out_prjparams = form_out,
        in_prjparams = forms_in,
        ...
    )
)
```

## Benefits
- Flexible per-edge form specification
- Maintains optimal performance where possible
- Graceful handling of form mismatches
- No breaking changes - all new parameters are optional

Let me know if any parts need clarification or expansion!


bvdmitri:
It seems to me that `ProjectionForm(MvNormalMeanCovariance, (2,), nothing)` basically copies the interface of `ProjectedTo(MvNormalMeanCovariance, 2, conditioner = nothing)`. Can we maybe use the `ProjectedTo` instead of introducing a new structure?

Nimrais:
> It seems to me that ProjectionForm(MvNormalMeanCovariance, (2,), nothing) basically copies the interface of ProjectedTo(MvNormalMeanCovariance, 2, conditioner = nothing). Can we maybe use the ProjectedTo instead of introducing a new structure? By doing that you can remove the create_project_to_ins and create_project_to entirely.

There are pros and cons to this suggestion.

The challenge arises from our design decisions in `ExponentialFamilyProjection.jl`, where we've tightly coupled `ProjectedTo` with `parameters`. This coupling makes the interface less flexible than ideal. If we were to remove `prjparams` from `CVIProjection` and use `ProjectedTo` only, then directly, we would be forced to specify form even in cases where we only want to specify the parameters and make the form changing during the inference time.

By keeping `ProjectionForm` separate, we maintain a cleaner separation of concerns:
- `ProjectionForm` handles just the form specification
- The projection parameters can be handled independently through `prjparams`

Or do you mean to have a union for `prjparams` where it could be `ExponentialFamilyProjection.ProjectionParameters` or `ProjectedTo`.

Nimrais:
>It seems to me that ProjectionForm(MvNormalMeanCovariance, (2,), nothing) basically copies the interface of ProjectedTo(MvNormalMeanCovariance, 2, conditioner = nothing). Can we maybe use the ProjectedTo instead of introducing a new structure? By doing that you can remove the create_project_to_ins and create_project_to entirely.

There are pros and cons.

The problem arises from our design decisions in `ExponentialFamilyProjection.jl`, where we've tightly coupled `ProjectedTo` with `parameters`. This coupling makes the interface less flexible than ideal. If we were to remove `prjparams` from `CVIProjection` and use `ProjectedTo` exclusively, we would be forced to specify the form even in cases where we only want to specify the parameters and allow the form to change dynamically during inference time.

bvdmitri:
>  This coupling makes the interface less flexible than ideal.

I don't agree with this, lets discuss in person. Having ability to pass `parameters` is even better. 

> we would be forced to specify the form even in cases where we only want to specify the parameters

Not really, you can always do `ProjectedTo(ExponentialFamilyDistribution, parameters = ...)` or simply dispatch on `ExponentialFamilyProjection.ProjectionParameters`. That's why I suggested you to simply reuse the `prjparams` field because its exactly what you want here. You duplicate structures that already exists and is supposed to solve the problem. Again lets better discuss in person.

ext/ReactiveMPProjectionExt/rules/marginals.jl
```diff
@@ -3,6 +3,39 @@ using TupleTools
...
+        conditioner = conditioner,
+        parameters =  ExponentialFamilyProjection.DefaultProjectionParameters()
+    )
+end
+
+function create_project_to_ins(::CVIProjection, form::ProjectedTo, ::Any)
+    return form
+end
+
+function create_project_to_ins(::CVIProjection, params::ProjectionParameters, ::Any)
```
bvdmitri, line 22: It looks to me that `m_in` ~(and other variables)~ is not really defined here

bvdmitri, line 22:
```suggestion
function create_project_to_ins(::CVIProjection, params::ProjectionParameters, m_in::Any)
```

ext/ReactiveMPProjectionExt/rules/out.jl
```diff
@@ -1,19 +1,33 @@
...
+    q_out_ef    = convert(ExponentialFamilyDistribution, q_out)
+    conditioner = getconditioner(q_out_ef)
+    return ProjectedTo(T, size(first(q_out_samples))...; conditioner = conditioner, parameters = ExponentialFamilyProjection.DefaultProjectionParameters())
+end
+
+function create_project_to(method::CVIProjection{R, S, OF}, ::Any, ::Any) where {R, S, OF <: ProjectedTo}
+    return method.out_prjparams 
+end
+
+function create_project_to(method::CVIProjection{R, S, OF}, ::Any, ::Any) where {R, S, OF <: ProjectionParameters}
```
bvdmitri, line 13: Here as well, variables haven't been defined

Nimrais: Yes I see sorry, I wonder why it does not fails in the tests

bvdmitri (approved):

Nimrais:
@bvdmitri codecoverage is working now! I see coverage for the project fails however is good for the patch. However the project codecoverage report is not informative for me.

bvdmitri:
Yeah afaik project coverage in Julia is stochastic, so it can fluctuate a bit, no worries. I approved the PR, you can merge, good job!

PR #435, opened 2024-12-18
fix #434: determine threads/cores via extra package Hwloc, if CpuId fails

### LINKED ISSUE #434: Parallelization in runtests.jl using ReTestItems and CpuId will not work for Apple M1 Pro CPU
I [noticed](https://github.com/m-j-w/CpuId.jl/issues/64) that CpuId yields default values when running `make test` on an Apple M1 Pro CPU:

```bash
$ julia --project=. -e "using CpuId; println(cpucores())"
0
```

This means that `CpuId`'s `cputhreads()` and `cpucores()` will return the default value `0` and consequently `ReTestItems`' `runtests` will run with only `1` thread and  `1` core.

To be able to run tests in parallel on an M1 chip, proposed solutions are

- add some special case for Apple M1 chips in `runtests.jl` before the issue in CpuId is fixed
- locate a different package that also returns threads/cores information for M1 chips 

The last commit to the `CpuId` package is over one year old, so the package seems to be unmaintained.

abpolym:
This PR should fix #434 .

Let me know if I need to add a test for this? It works now locally and Hwloc only triggers if CpuId fails (which is only in the case of ARM machines).

As soon as this works, I'll update the runtests.jl in the other packages accordingly.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,10 +1,8 @@
-using Aqua, CpuId, ReTestItems, ReactiveMP, Hwloc
+using Aqua, ReTestItems, ReactiveMP, Hwloc
 
 Aqua.test_all(ReactiveMP; ambiguities = false, piracies = false, deps_compat = (; check_extras = false, check_weakdeps = true))
 
-nthreads, ncores = cputhreads(), cpucores()
-nthreads = nthreads == 0 ? Hwloc.num_virtual_cores() : nthreads
-ncores = ncores == 0 ? Hwloc.num_physical_cores() : ncores
+nthreads, ncores = Hwloc.num_virtual_cores(), Hwloc.num_physical_cores()
 nthreads, ncores = max(nthreads, 1), max(ncores, 1)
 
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -1,8 +1,10 @@
-using Aqua, CpuId, ReTestItems, ReactiveMP
+using Aqua, CpuId, ReTestItems, ReactiveMP, Hwloc
```
bvdmitri, line 1: Do we need CpuId then? Can we just use `Hwloc` in all situations?

abpolym: I think the only argument for `CpuId` is that it is very fast since you only have to run one instruction `cpuid`.
We only need it once so the theoretical performance increase is non-existant or minimal at best 😄.

Should I just remove the CpuId requirement 👍 👎 ?

abpolym:
Removed dependency for `CpuId`, `runtests.jl` only uses `Hwloc` now.

bvdmitri (approved):

PR #442, opened 2025-01-24
Rule updates

wouterwln:
These are some QoL updates for rules. I'll keep this a draft until I have some more rules respecified.

### DIFF
--- a/src/rules/normal_mixture/m.jl
+++ b/src/rules/normal_mixture/m.jl
@@ -1,6 +1,6 @@
 export rule
 
-@rule NormalMixture((:m, k), Marginalisation) (q_out::Any, q_switch::Any, q_p::Any) = begin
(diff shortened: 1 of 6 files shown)


src/rules/normal_mixture/m.jl
```diff
@@ -1,6 +1,6 @@
 export rule
 
-@rule NormalMixture((:m, k), Marginalisation) (q_out::Any, q_switch::Any, q_p::Any) = begin
+@rule NormalMixture((:m, k), Marginalisation) (q_out::Any, q_switch::Union{Categorical, PointMass{<:Vector}}, q_p::Any) = begin
```
bvdmitri, line 3:
```suggestion
@rule NormalMixture((:m, k), Marginalisation) (q_out::Any, q_switch::Union{Categorical, Bernoulli, PointMass{<:AbstractVector}}, q_p::Any) = begin
```

bvdmitri (changes requested):
Tests are failing because `Bernoulli` is also allowed in the switch

bvdmitri (approved):

PR #447, opened 2025-02-04
Small addition to BinomialPolya

ismailsenoz:
BinomialPolya node messages for $\beta$ interface currently works with multivariate Normals. This PR extends it to work with univariate normal for binary regression.


bvdmitri:
Nice @ismailsenoz , I pushed a change to your code to use a proper API from BayesBase for this. There are no tests in PR, so I couldn't double check. Could you double check that my changes work on your side?

ismailsenoz:
Thanks @bvdmitri  ! I forgot to push the test for 1d case in the original PR. Your addition works. I merged with main and pushed my changes.

bvdmitri (approved):

PR #453, opened 2025-02-20
Softdot doc

Raphael-Tresor:
The docstring for the SofDot to know which variable represents what.


bvdmitri (approved):
Thanks @Raphael-Tresor , I adjusted the wording a bit since "multiplication" and "dot" are different operations, so the original documentation wasn't entirely correct. I fixed that. Thanks for the PR!

Raphael-Tresor:
thanks for the review  @bvdmitri  :)

PR #456, opened 2025-03-05
fix: softdot average energy

ismailsenoz:
In the current average energy computation for the softdot node under structured factorization assumption the indices in the following line `ar_slice(F, myx, (order + 1):(2order)), ar_slice(F, Vyx, (order + 1):(2order), (order + 1):(2order))` are inaccurate. Suppose x is 2d and \theta is also 2d. Then y is 1d. Meaning order is 2 in this case. In this example then the slice will range between 3:4 causing bounds error. This PR attempts to resolve this issue.

### DIFF
--- a/src/nodes/predefined/softdot.jl
+++ b/src/nodes/predefined/softdot.jl
@@ -2,27 +2,7 @@ export softdot, SoftDot
 
 import StatsFuns: log2π
 
-"""
-SoftDot [aliases = [softdot]]: ReactiveMP node for message passing. 
-
-The SoftDot node can be used as a substitute for the dot product operator delta node (the outgoing variable is the dot product of two others). It softens the delta constraint by adding a Gaussian noise as follows:
-
-    y ~ N(dot(θ, x), γ^(-1))
-
-Interfaces:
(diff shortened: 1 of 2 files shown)


bvdmitri:
> slice will range between 3:6 causing bounds error
am I missing something or it should be `3:4`? because as you mentioned `order=2` and the code has `2order` which is `4`?

ismailsenoz:
I think my suggestion is also not correct @albertpod I agree. Do you have a fix in mind?

ismailsenoz:
`mx, Vx = ar_slice(F, myx, (2):(order + 1)), ar_slice(F, Vyx, (2):(order + 1), (2):(order + 1))` this is also what I suggested @albertpod

albertpod:
@ismailsenoz I think your fix is right, but in the test we need to do:
`q_y_x = MvNormalMeanCovariance(zeros(3), diageye(3))`

albertpod:
Imo this is okee, now. Unless no objections, I would merge it once tests pass.

bvdmitri (approved):

PR #484, opened 2025-07-04
A lesson in numerical stability

wouterwln:
Softmaxing apparently helps a lot with numerical stability


bvdmitri (approved):
This is a very nice catch :D

PR #502, opened 2025-09-09
Optimize CTransition rules

### LINKED ISSUE #368: The computations of rules for ContinuousTransition node can be improved
Following the discussions during [PR](https://github.com/biaslab/ReactiveMP.jl/pull/309), @bartvanerp has identified a few places where the computations around the `ContinuousTransition` node can be improved.
In particular
1. If the transformation of RV `a` is linear, then the computation of Jacobians can be done only once.
2. There are multiple places where trace operation within `ContinuousTransition` can be optimized, e.g., in the computation of the average energy.

bvdmitri:
A PR to check if tests are failing or not

### DIFF
--- a/src/helpers/algebra/common.jl
+++ b/src/helpers/algebra/common.jl
@@ -99,7 +99,7 @@ function mul_trace(A::AbstractMatrix, B::AbstractMatrix)
     n = first(sA)
     for i in 1:n
         for j in 1:n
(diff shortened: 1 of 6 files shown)


albertpod:
This is a very useful node, I look forward @ReactiveBayes/biaslab improving the scalability.

Issue #432, opened 2024-12-11
Categorical node expects one-hot encoded vector, but doesn't check it


bvdmitri:
The rules should check it and throw a user friendly error

abpolym:
I will fix this

abpolym:
Yeah I forgot about this issue, lets do it during the dev session

PR #510, opened 2025-10-02
#432: Check if probvec of Categorical q_out is one-hot encoded vector 

abpolym:
I assume the issue only relates to the `p` rule; to check if the input (`q_out`)  is a one-hot encoded vector.
I introduced two helper functions - one to check if a vector is one-hot encoded, and another one that moved to functionality of both rules (which were almost the same except a `@logscale` macro call for the `PointMass` case). However, I had to stay with a separate logic for the `PointMass` case since I got the following error: `UndefVarError: "_addons" not defined`


bvdmitri:
Thanks @abpolym ! I will add it to my TODO

bartvanerp (approved):

PR #531, opened 2025-10-21
implement missing marginal rule multiplication node and added tests

RetrospectiveRotations:
In marginals.jl, I added the following rule:

# Specific version for univariate and pointmass input
@marginalrule typeof(*)(:A_in) (m_out::MvNormalMeanPrecision, m_A::UnivariateNormalDistributionsFamily, m_in::PointMass{<:AbstractVector}, meta::Any) = begin 
    m_outbound_A = @call_rule typeof(*)(:A, Marginalisation) (m_out=m_out, m_in=m_in, meta=meta)
    q_a = prod(ClosedProd(), m_A, m_outbound_A)
    return (A = q_a, in = m_in)
end

Furthermore, I added marginals_test.jl which tests that marginal rule & the other 2 marginal rules that are already implemented for 3 parameter configurations.

### DIFF
--- a/src/rules/multiplication/marginals.jl
+++ b/src/rules/multiplication/marginals.jl
@@ -21,4 +21,3 @@ end
     q_a = prod(ClosedProd(), m_A, m_outbound_A)
     return (A = q_a, in = m_in)
 end
-


src/rules/multiplication/marginals.jl
```diff
@@ -11,5 +11,13 @@ end
 @marginalrule typeof(*)(:A_in) (
     m_out::UnivariateNormalDistributionsFamily, m_A::UnivariateNormalDistributionsFamily, m_in::PointMass{<:Real}, meta::Union{<:AbstractCorrectionStrategy, Nothing}
 ) = begin
-    return @call_marginalrule typeof(*)(:A_in) (m_out = m_out, m_A = m_in, m_in = m_A, meta = meta)
+    flipped_result = @call_marginalrule typeof(*)(:A_in) (m_out = m_out, m_A = m_in, m_in = m_A, meta = meta)
+    return (A = flipped_result[:in], in = flipped_result[:A])
```
wouterwln: Good catch @RetrospectiveRotations !!

PR #540, opened 2025-10-27
Robust invS in outgoing Λ-rule of MvNormalMeanPrecision-node

skoghoern:
I stumbled upon this small edge case during inference. check if you think it is helpful, inference also worked without it, but i thought it might add more stability.
### Description
when the incoming messages ($q_{\text{out}}$ and $q_{\mu}$) are identical point masses, the invS became a singularity, **zero matrix**, leading to an **undefined Wishart distribution**.
To prevent this, this PR introduces a small regularization term ($\epsilon \mathbf{I}$) to the calculation of the inverse scale matrix ($\mathbf{invS}$) in the $\mathbf{\Lambda}$ VMP update.
### Changes

The `invS` calculation now adds $10^{-6}$ times the identity matrix ($\mathbf{I}$) for robustness.

```julia
invS_raw = v_mean + v_out + (m_mean - m_out) * (m_mean - m_out)'
invS = invS_raw + 1e-6 * diagm(ones(D))
```

This regularization **guarantees $\mathbf{invS}$ is positive definite**, ensuring the resulting Wishart distribution for the precision matrix is always proper.

### DIFF
--- a/src/rules/mv_normal_mean_precision/precision.jl
+++ b/src/rules/mv_normal_mean_precision/precision.jl
@@ -1,25 +1,24 @@
 # Variational                       # 
 # --------------------------------- #
-@rule MvNormalMeanPrecision(:Λ, Marginalisation) (q_out::Any, q_μ::Any) = begin
+@rule MvNormalMeanPrecision(:Λ, Marginalisation) (q_out::Any, q_μ::Any, meta::Union{<:AbstractCorrectionStrategy, Nothing}) = begin
     m_out, v_out   = mean_cov(q_out)
     m_mean, v_mean = mean_cov(q_μ)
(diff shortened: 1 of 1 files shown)


bvdmitri:
I think its a very nice change! But I would rather use our [MatrixCorrectionTools](https://github.com/ReactiveBayes/MatrixCorrectionTools.jl) library for this purpose. Look at the example [here](https://github.com/ReactiveBayes/ReactiveMP.jl/blob/8a0be859bdf78f88c103fd2a5dbd127c7d3e11db/src/rules/multiplication/in.jl#L9). The idea here is that you can use the `meta` to allow for different strategies (e.g. maybe `1e-6` is not enough)

Issue #294, opened 2023-03-06
`@rule`, `@marginalrule` and `@average_energy` should check interface names


bvdmitri:
Together with @MagnusKoudahl we spent quite a significant amount of time just to realize that we misspelled `q_A::Any` and `q_a::Any` for some of the message update rules. It would be nice if `@rule` and friends would throw an error if one attempts to create a rule with misspelled input arguments.

abpolym:
This sounds pretty easy to do. Any progress since the last 2.5 years or can I take over? 😄

abpolym:
Alright I have added functionality that throws an error for `@rule` and `@marginalrule` if the names of the interfaces in the `@node` definition does not match with the ones provided in the `@rule` statement (covering also cases like Mixtures with switch indices, structural messages for joints like q_y_x etc).


It breaks the `basic test_rule macro invokation` test in test/rule_tests.jl:rule (see `make test test_args="rule_tests.jl"`), because `LoadError: UndefVarError: `TestNodeForTestRuleMacro` not defined`.

@bvdmitri could you have a quick look over that?

We also need to add the check for the `@average_energy` macro and add some tests.

Right now it will throw an error if we e.g. misname interface names:

```julia
using ReactiveMP, BayesBase, Distributions, ExponentialFamily
@rule GammaMixture((:a, k), Marginalisation) (q_out::Any, q_switche::Any, q_b::GammaDistributionsFamily) = nothing

ERROR: LoadError: ArgumentError: Rule interface mismatch for @rule GammaMixture((:a, k), Marginalisation) (q_out::Any, q_switche::Any, q_b::GammaDistributionsFamily) = nothing:
  Expected symbols: a, b, out, switch
  Provided symbols: a, b, out, switche
  Missing symbols:  switch

Stacktrace: [...]
```

Note that it does not throw an error (yet) if we define extra interfaces:

```julia
@rule GammaMixture((:a, k), Marginalisation) (q_out::Any, q_switch::Any, q_b::GammaDistributionsFamily, q_extra::Any) = nothing
```

It probably should throw an error?

abpolym:
The `check_rule_interfaces` function now also checks for extra interfaces and throws an appropriate error:

```julia
ERROR: LoadError: ArgumentError: Rule interface mismatch for @rule GammaMixture((:a, k), Marginalisation) (q_out::Any, q_switch::Any, q_b::GammaDistributionsFamily, q_extra::Any) = nothing:
[...]

PR #545, opened 2025-11-01
294 rule marginalrule and average energy should check interface names

abpolym:
Relevant issue: #294

I would like to get a first review on the code.
The code now checks for misnamed interfaces, including extra interfaces etc (see issue).

Potential things to do:
* Aliases are not considered right now
* Tests are missing
* `basic test_rule macro invokation` test in `test_rules.jl` is breaking due to specification of a node that is unknown when the rule for that node is being called


bvdmitri (commented):
I still have some idea for this PR so lets not merge just yet

PR #550, opened 2025-11-11
#549 Add MessageMapping docs

### LINKED ISSUE #549: What is message mapping?
I'm working on understanding and improving `AddonMemory`, but I don't understand what "message mapping" is, and there are do comments or docstrings. Could someone document this in the code?

abpolym:
Should resolve #549 . I also added `LiveServer` as a dependency to `docs/Project.toml`.

The docs now should show this:

Let me know if the docs bla bla that I added is actually correct haha


bvdmitri:
Tests for 1.10 are failing is unrelated so I'll merge

PR #582, opened 2026-03-16
Make the formatting rules stricter

bvdmitri:
I think we did a mistake when set the margin to a very large value of `180`. It makes it harder to read code in multi-tab setting, lines are just way too long. It also makes it harder with an extra AI panel open on the side. I suggest we make it stricter and reduce the margin from `180` to `80` to make it multi-tab friendly. 

@Nimrais @wouterwln wdyt? I think its worth it, if you agree I make the same for RxInfer, but lets test here first

This PR simply changes the formatting options, all the other changes are due to `make format`


Nimrais (approved):
Shorter lines better, so I approve.

PR #589, opened 2026-03-24
Trace ID to link before and after events

fonsp:
I think it is useful to link before and after events. Here is one implementation, using a field `trace_id` added to events. This also needs to be done to the events defined in RxInfer.jl.

It works, but maybe we can do better?

# Alternative 1: `before` field in `AfterSomething` events
We can add a field `.before` that stores the counterpart of an `AfterSomething` event. Instead of `trace_id`. (Then the 'before' objectID functions the trace ID.)

# Aternative 2: `Before{...}` and `After{...}` types
We could change the types to eg:

```
Before{ProductOfMessagesEvent{V, C, Ms, R}} <: After{Event{:product_of_messages}}
```

The `After` type can have a `.result` field.

Then we can replace this:

```julia
# Before
trace_id = uuid4()
invoke_callback(
    context.callbacks,
    BeforeProductOfMessagesEvent(variable, context, messages, trace_id),
)

result = as_message(...)

invoke_callback(
    context.callbacks,
    AfterProductOfMessagesEvent(variable, context, messages, trace_id),
)
```

with:

```julia
# After
result = with_span(
	BeforeProductOfMessagesEvent(variable, context, messages, trace_id)
) do
	as_message(...)
end
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ReactiveMP"
 uuid = "a194aa59-28ba-4574-a09c-4a745416d6e3"
-version = "5.6.6"
 authors = ["Dmitry Bagaev <d.v.bagaev@tue.nl>", "Albert Podusenko <a.podusenko@tue.nl>", "Bart van Erp <b.v.erp@tue.nl>", "Ismail Senoz <i.senoz@tue.nl>"]
+version = "5.6.6"
 
 [deps]
 BayesBase = "b4ee3484-f114-42fe-b91c-797d54a0c67e"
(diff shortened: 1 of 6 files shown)


wouterwln (approved):
Super nice, I like this a lot

bvdmitri:
It seems that Julia v1.10 isn't that smart :( I'm ok with just guarding the test with `@static VERSION` check and only check on v1.11+

bvdmitri:
OR - it might also happening because `sqrt` can throw, an maybe Julia v1.10 isn't good enough to understand that and shows allocations for potential error path. Its better to call functions that truly do not allocate, e.g. `+`

bvdmitri (approved):
lets gooo!

fonsp:
I'll work through the CI Julia 1.10 and then I merge 👍

fonsp:
For future reference, here is a function that matches traces based on their ID:

```julia
function find_spans(trace_events::Vector{TracedEvent})
	# Found spans
	spans = Tuple{TracedEvent,TracedEvent}[]
	# Currently waiting to be matched
	lonely = OrderedDict{Base.UUID, Any}()
	# No match possible
	not_matched = TracedEvent[]

	for te in trace_events
		event = te.event
		if hasfield(typeof(event), :trace_id)
			tid = event.trace_id
			friend = get(lonely, tid, nothing)
			if !isnothing(friend)
				delete!(lonely, tid)
				push!(spans, (friend, te))
			else
				lonely[tid] = te
			end
		else
			push!(not_matched, te)
		end
	end

	push!(not_matched, values(lonely)...)

	return (; spans, not_matched )
end
```

PR #595, opened 2026-04-02
Better trace ids

bvdmitri:
This PR changes the generation of trace ids (also renamed to span ids) in order to remove massive performance hit for certain models. In my benchmark, the generation of trace IDs affected benchmarks by a factor of 7! And this was even without enabling the callbacks. The current PR changes the implementation a bit by calling an auxiliary function `ReactiveMP.generate_span_id` that would simply return `nothing` if callbacks are also `nothing`. That fixes the performance hit in case when callbacks are not set, but otherwise it still uses `uuid4` (definitely **the** culprit). We may want to use a simple global Int counter for that? Not sure if it would be faster though since this counter would need to be guarded with semaphores to protect from two threads generating the same span ID. 

@fonsp wdyt?

### DIFF
--- a/docs/src/lib/callbacks.md
+++ b/docs/src/lib/callbacks.md
@@ -38,6 +38,21 @@ To see which fields an event carries, use the standard Julia introspection:
 julia> ?ReactiveMP.BeforeProductOfTwoMessagesEvent
 ```
 
+## Event spans
+
+Certain events create a "span". For example all "before" and "after" events 
+can be considered together. To track these relationships ReactiveMP uses the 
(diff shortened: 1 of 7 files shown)


bvdmitri:
The affected benchmark is this one just in case 
Screenshot 2026-04-02 at 14 55 11

fonsp (approved):
Beautiful!

bvdmitri:
@fonsp could you adjust `trace_id` -> `span_id` in RxInfer accordingly? I just noticed we do there as well, basically just need to change the fields and use the `ReactiveMP.generate_span_id` there too

bvdmitri:
https://github.com/ReactiveBayes/RxInfer.jl/pull/634

PR #600, opened 2026-04-17
Release 6

bvdmitri:
# Summary

ReactiveMP.jl v6.0 — a major release that overhauls the pipeline/addon system, introduces callbacks, annotations, and a modernized documentation & branding.

## Highlights
- New Stream Postprocessor Abstraction
- Replaces the old AbstractPipelineStage hierarchy with AbstractStreamPostprocessor, unifying outbound message, marginal, and score stream postprocessing under a single concept
- ScheduleOnStreamPostprocessor supersedes ScheduleOnPipelineStage and per-node schedulers
- CompositeStreamPostprocessor for chaining postprocessors
Callback/Event System
- Hook into message passing steps: rule calls, message products, form constraints, marginal computation
- MessageProductContext bundles product computation settings and callbacks
Annotations System (replaces Addons)
- AnnotationDict, LogScaleAnnotations, InputArgumentsAnnotations replace AddonLogScale, AddonMemory, AddonDebug
- @logscale macro for setting log-scale annotations inside @rule bodies
- getannotations replaces getaddons; Message/Marginal type parameters simplified
API Renames
- messagein/messageout → get_stream_of_inbound_messages/get_stream_of_outbound_messages
- getmarginal/getprediction → get_stream_of_marginals/get_stream_of_predictions
- setmarginal!/setmessage! → set_initial_marginal!/set_initial_message!
- update! → new_observation!
- MarginalSkipStrategy hierarchy → skip_initial(), skip_clamped(), skip_clamped_and_initial() filter operators
Documentation & Branding
- New concepts pages: factor graphs, message passing, reactive programming, inference lifecycle
- Expanded library documentation for variables, nodes, messages, marginals
- New documentation pages for stream postprocessors and callbacks
- Migration guide for v5 → v6
- Project logo and updated README

## Other
- Variable labels for RandomVariable, ConstVariable, DataVariable
- create_new_stream_of_inbound_messages! for per-connection message streams
- get_stream_of_predictions/set_stream_of_predictions! accessors
- Switched from ReTestItems to TestItemRunner

PR #605, opened 2026-05-18
fix: improve AddonMemory deprecation error message with migration example

ArpanC6:
### What

Made the AddonMemory deprecation error message in
src/annotations/input_arguments.jl easier to understand.

Now it shows:

A simple example of what to change before/after
A direct link to the v5-to-v6 migration guide

### Why

Before, the error message was confusing and didn’t tell users what to do (like in RxInfer.jl issue #600).
Now it clearly shows what code to replace and where to find help.

### Changes
src/annotations/input_arguments.jl: updated the error message

1)  Tests pass locally
2)  Code formatted
3) CHANGELOG.md updated

### DIFF
--- a/src/annotations/input_arguments.jl
+++ b/src/annotations/input_arguments.jl
@@ -183,7 +183,7 @@ function AddonMemory(args...; kwargs...)
         """To migrate, replace:\n""" *
         """  addons = (AddonMemory(),)\n""" *
         """with:\n""" *
-        """  addons = (InputArgumentsAnnotations(),)\n""" *
+        """  annotations = (InputArgumentsAnnotations(),)\n""" *
         """See the migration guide: https://reactivebayes.github.io/ReactiveMP.jl/stable/migration-guides/v5-to-v6/""",
     )
 end
--- a/src/annotations/input_arguments.jl
+++ b/src/annotations/input_arguments.jl
@@ -178,7 +178,12 @@ Use [`InputArgumentsAnnotations`](@ref) instead. See the migration guide in the 
 """
 function AddonMemory(args...; kwargs...)
     error(
-        """`AddonMemory` has been removed in ReactiveMP v6 and replaced by `InputArgumentsAnnotations`. """ *
-        """See the migration guide in the documentation for details.""",
+        """`AddonMemory` has been removed in ReactiveMP v6 """ *
+        """and replaced by `InputArgumentsAnnotations`.\n""" *
+        """To migrate, replace:\n""" *
+        """  addons = (AddonMemory(),)\n""" *
+        """with:\n""" *
+        """  addons = (InputArgumentsAnnotations(),)\n""" *
+        """See the migration guide: https://reactivebayes.github.io/ReactiveMP.jl/stable/migration-guides/v5-to-v6/""",
     )
 end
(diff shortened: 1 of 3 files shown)


src/annotations/input_arguments.jl
```diff
@@ -178,7 +178,12 @@ Use [`InputArgumentsAnnotations`](@ref) instead. See the migration guide in the
...
 function AddonMemory(args...; kwargs...)
     error(
-        """`AddonMemory` has been removed in ReactiveMP v6 and replaced by `InputArgumentsAnnotations`. """ *
-        """See the migration guide in the documentation for details.""",
+        """`AddonMemory` has been removed in ReactiveMP v6 """ *
+        """and replaced by `InputArgumentsAnnotations`.\n""" *
+        """To migrate, replace:\n""" *
+        """  addons = (AddonMemory(),)\n""" *
+        """with:\n""" *
+        """  addons = (InputArgumentsAnnotations(),)\n""" *
```
bvdmitri:
```suggestion
        """  annotations = (InputArgumentsAnnotations(),)\n""" *
```

bvdmitri (changes requested):
maybe also use the `annotations` as a variable name instead of `addons`

ArpanC6:
Hlw @bvdmitri I have improved the `AddonMemory` deprecation 
error message to include a clear before/after migration 
example and a direct link to the v5-to-v6 migration guide.

This should help users like in RxInfer.jl#600 who were 
confused about how to migrate.

ArpanC6:
@bvdmitri thanks for the review..

While working on this I noticed the error message 
for AddonLogScale in the same file follows a 
similar pattern should I improve that one too 
for consistency? Alsoare there other deprecated functions in the 
codebase that might benefit from similar migration 
examples?

bvdmitri:
> While working on this I noticed the error message
for AddonLogScale in the same file follows a
similar pattern should I improve that one too
for consistency?

Would be great! 

> Alsoare there other deprecated functions in the
codebase that might benefit from similar migration
examples?

Not that I immediately can remember :) I think we can focus this PR on AddonLogScale

src/annotations/logscale.jl
```diff
@@ -80,7 +80,13 @@ Use [`LogScaleAnnotations`](@ref) instead. See the migration guide in the docume
 """
 function AddonLogScale(args...; kwargs...)
     error(
-        """`AddonLogScale` has been removed in ReactiveMP v6 and replaced by `LogScaleAnnotations`. """ *
-        """See the migration guide in the documentation for details.""",
+        """`AddonLogScale` has been removed in ReactiveMP v6 """ *
+        """and replaced by `LogScaleAnnotations`.\n""" *
+        """\n""" *
+        """To migrate, replace:\n""" *
+        """  annotations = (AddonLogScale(),)\n""" *
```
bvdmitri: Here is the opposite, previously we called it `addons`, so the variable in the old code is most likely to be called `addons`


```suggestion
        """  addons = (AddonLogScale(),)\n""" *
```

bvdmitri (changes requested):

ArpanC6:
Updated Used addons as the variable name in the AddonLogScale migration example for consistency with the old API.

bvdmitri:
Thanks @ArpanC6 !

ArpanC6:
> Thanks @ArpanC6 !

Welcome..

PR #612, opened 2026-06-17
feat(ar): ConjugateAR node with structured MvNormalGamma (θ, γ) posterior

RetrospectiveRotations:
## Summary

Adds `ConjugateAR`, an autoregressive node that carries the AR coefficients `θ` and the
transition precision `γ` jointly on a single edge `w` as an `MvNormalGamma`, instead of the
mean-field `q(θ)q(γ)` split of the existing `AR` node. Because `(θ, γ)` are jointly conjugate
for the AR likelihood (a Bayesian linear regression with unknown noise precision), the
structured posterior `q(θ, γ)` is exactly `MvNormalGamma` — more accurate than mean-field and
still closed-form.

Also registers `MvNormalGamma` as a prior factor node so models can write
`w ~ MvNormalGamma(μ, Λ, α, β)`.

## ⚠️ Dependency

This branch depends on **ReactiveBayes/ExponentialFamily.jl#287** (the `MvNormalGamma`
distribution + `prod` rule). That PR must land first; this one should not be merged until the
ExponentialFamily release containing `MvNormalGamma` is available.

## What's included

- **`ConjugateAR` node** `[y, x, w]` reusing `ARMeta`.
- **`ConjugateAR(:w)` marginal rule** — the structured BLR Normal-Gamma posterior, with
  expected sufficient statistics from `q(y, x)`: `C = E[xxᵀ]`, `b = E[x·y₁]`, `a = E[y₁²]`.
- **`ConjugateAR(:w)` message rule** — the AR likelihood as a Normal-Gamma factor; its product
  with the prior at the equality node reproduces the `:w` marginal (natural parameters add).
- **`ConjugateAR(:x)`, `(:y)`, `(:y_x)`** state/marginal rules and **`@average_energy`**,
  delegating to the tested `AR` rules via effective `(q_θ, q_γ)` derived from `q(w)` (no
  duplicated algebra).
- **`MvNormalGamma` prior node** `[out, μ, Λ, α, β]` with its `:out` message rule and
  `@average_energy` (so `free_energy = true` works for `ConjugateAR` models).

## Math references

- Normal-Gamma density — https://statproofbook.github.io/P/ng-pdf
- BLR Normal-Gamma posterior — https://statproofbook.github.io/P/blr-post
- Normal-Gamma differential entropy — https://statproofbook.github.io/P/ng-dent

All formulas were cross-checked against these references and for internal consistency
(message ↔ marginal ↔ `prod`, entropy ↔ self cross-entropy, univariate ↔ `NormalGamma`).

## Test plan

- [x] `ConjugateAR(:w)` message params + `prod(prior, message) == :w` marginal
- [x] `(:x)`, `(:y)`, `(:y_x)` delegation matches `AR` with effective `(q_θ, q_γ)`
- [x] BLR reference (orders 1–3), univariate `NormalGamma` reduction, properness/PSD
- [x] `MvNormalGamma(:out)` pass-through; average energy (self cross-entropy == entropy)
- [x] End-to-end RxInfer AR(2): recovers transition precision (0.99 vs 1.0) and coefficients
- [ ] CI green on this branch


wmkouw (approved):
Looks great. Go for it.

My only comment is that the conjugate_autoregressive/y.jl message can already be cast as a Tdist. I realized _after_ our call that the reason we don't have T-distributions in ExponentialFamily.jl is because T-distributions are not part of the exponential family. They do exist in Distributions.jl (https://juliastats.org/Distributions.jl/stable/univariate/#Distributions.TDist; although it may not be full location-scale parameterization).

RetrospectiveRotations:
> Looks great. Go for it.
> 
> My only comment is that the conjugate_autoregressive/y.jl message can already be cast as a Tdist. I realized _after_ our call that the reason we don't have T-distributions in ExponentialFamily.jl is because T-distributions are not part of the exponential family. They do exist in Distributions.jl (https://juliastats.org/Distributions.jl/stable/univariate/#Distributions.TDist; although it may not be full location-scale parameterization).

I asked our mutual friend for the reason of the current implementation. Since TDist isn't exp-family it has no closed-form product with the downstream Gaussian message, so as an internal message we'd just project back to Gaussian, right? Do you want me to change this?

wmkouw:
Well, the `y` message is for output predictions which tend not to be re-used. And there are multiple ways of doing the Gaussian approximation (e.g., mean-field VI for accurate mean, moment matching for accurate variance).

But this is fine for now. If we need more, we can always open up a new branch with some options under the ARmeta.

bvdmitri (approved):

PR #616, opened 2026-07-09
Fix spurious extra mean(q_γ) in softdot mean-field average energy

### LINKED ISSUE #615: softdot mean-field @average_energy has a spurious extra mean(q_γ) on the cross term
### Description

The mean-field `@average_energy` for `softdot` scales the cross term by `mean(q_γ)` twice, so the average energy is wrong whenever `mean(q_γ) ≠ 1`.

In `src/nodes/predefined/softdot.jl` (present unchanged from at least v4.4.3 through v6.3.1 and current `main`):

```julia
@average_energy softdot (q_y::Any, q_θ::Any, q_x::Any, q_γ::Any) = begin
    ...
    return (-mean(log, q_γ) + log2π + m_γ * (V_y + m_y^2 - 2m_γ * m_y * m_θ'm_x + ...)) / 2
    #                                        ^^^^ extra m_γ: the whole bracket is already
    #                                             multiplied by m_γ, so the cross term
    #                                             gets m_γ² instead of m_γ
end
```

The correct energy is

```
U = ½( log 2π − ⟨log γ⟩ + ⟨γ⟩·⟨(y − θᵀx)²⟩ ),
⟨(y − θᵀx)²⟩ = V_y + m_y² − 2·m_y·m_θᵀm_x + tr(V_θ V_x) + m_xᵀV_θ m_x + m_θᵀ(V_x + m_x m_xᵀ)m_θ
```

i.e. the cross term should be `- 2 * m_y * m_θ'm_x` with no inner `m_γ`.

The **structured** variant in the same file is correct (`- 2 * mθ' * (Vy1x + mx * my1)`, no inner `mγ`), so the two variants disagree for any `mean(q_γ) ≠ 1`.

### Reproducer

```julia
using ReactiveMP, ExponentialFamily, BayesBase, StatsFuns

q_y = NormalMeanVariance(0.4, 0.6)
q_θ = NormalMeanVariance(1.2, 0.25)
q_x = NormalMeanVariance(0.5, 0.3)
q_γ = GammaShapeRate(4.0, 1.0)     # mean(q_γ) = 4

marg = map(q -> Marginal(q, false, false, nothing), (q_y, q_θ, q_x, q_γ))
ae = score(AverageEnergy(), softdot, Val{(:y, :θ, :x, :γ)}(), marg, nothing)

# hand-computed reference
S = 0.6 + 0.4^2 - 2*0.4*1.2*0.5 + 0.25*0.3 + 0.5^2*0.25 + 1.2^2*(0.3 + 0.5^2)
ref = 0.5 * (log2π - (digamma(4.0) - log(1.0)) + 4.0 * S)

ae - ref   # ≠ 0; equals ½·mγ·(mγ−1)·(−2·m_y·m_θᵀm_x) = ½·4·3·(−0.48) → −2.88 off
```

### Impact

Message rules are unaffected (they are correct); only the mean-field Bethe free energy contribution of `softdot` is biased. Any workflow reading `free_energy = true` with a `softdot` node under full mean-field factorization gets a wrong BFE value, and the sign/size of the error varies with `E[γ]`, which matters for FE-based model comparison, convergence monitoring, and FE-driven adaptive schemes.

### Suggested fix

```julia
return (-mean(log, q_γ) + log2π + m_γ * (V_y + m_y^2 - 2 * m_y * m_θ'm_x + mul_trace(V_θ, V_x) + m_x'V_θ * m_x + m_θ' * (V_x + m_x * m_x') * m_θ)) / 2
```

which makes the mean-field variant consistent with the structured `q_y_x` variant in the same file.

Found while wrapping `softdot` for a tempered-inference project (RxInfer 4.7.3 / ReactiveMP 5.6.6); verified the term against a hand derivation and the structured variant.

bvdmitri:
Closes #615.

## Problem

The mean-field `@average_energy` for the `softdot` node (`src/nodes/predefined/softdot.jl`) multiplied the whole `⟨(y − θᵀx)²⟩` bracket by `mean(q_γ)` and **also** carried an extra inner `mean(q_γ)` on the cross term:

```julia
m_γ * (
    V_y + m_y^2 - 2m_γ * m_y * m_θ'm_x +   # ← inner m_γ is wrong
    ...
)
```

so the cross term was scaled by `mean(q_γ)²` instead of `mean(q_γ)`. The correct energy is

```
U = ½( log2π − ⟨log γ⟩ + ⟨γ⟩·⟨(y − θᵀx)²⟩ )
```

with a single power of `⟨γ⟩` multiplying the entire bracket. The bias vanishes only when `mean(q_γ) = 1`.

## Fix

Drop the inner `m_γ` on the cross term. This makes the mean-field variant consistent with the (already-correct) structured `q(y, x)` variant in the same file, whose cross term is `−2 * mθ' * (Vy1x + mx * my1)` with no inner `mγ`.

## Impact

- **Message rules are unaffected** — they were and remain correct, so inference results/posteriors do not change.
- Only the Bethe free energy contribution of a `softdot` node under **full mean-field factorization** was wrong, and only for `mean(q_γ) ≠ 1`.
- The error was `½·mean(q_γ)·(mean(q_γ)−1)·(−2·m_y·m_θᵀm_x)`, growing with `E[γ]`. This matters for `free_energy = true` workflows: FE-based model comparison, convergence monitoring, and FE-driven adaptive schemes.

## Provenance

The bug was present unchanged from at least v4.4.3 through v6.3.1, so any Bethe free energy values previously reported for mean-field `softdot` models were biased. Flagged in the CHANGELOG. Originally found by @mhidalgoaraya while wrapping `softdot` for a tempered-inference project (#615).

## Tests

The two pre-existing mean-field tests used hardcoded expected values that **encoded the buggy behaviour**, so they were replaced with self-checking regressions:

- **Closed-form reference** — mean-field score vs. an explicit `U = ½(log2π − ⟨log γ⟩ + ⟨γ⟩·⟨(y − θᵀx)²⟩)` reference across several `mean(q_γ)` values (uni- and multivariate).
- **Hand-computed value (#615 reproducer)** — the exact reproducer from the issue, which the buggy code missed by `−2.88`.
- **Mean-field ↔ structured consistency** — with zero posterior `y`–`x` covariance the two `@average_energy` methods must agree for any `mean(q_γ)`; they disagreed under the bug. Covers uni- and multivariate `x`.

The structured-variant magic-number tests are unchanged (that path is unaffected).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,7 @@ jobs:
   test:
     name: Tests ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
     continue-on-error: ${{ contains(matrix.version, 'nightly') }}
     permissions:
       actions: write
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Mean-field `@average_energy` for the `softdot` node no longer applies a spurious extra `mean(q_γ)` factor to the cross term `−2·m_y·m_θᵀm_x`, which biased the term by `mean(q_γ)²` instead of `mean(q_γ)`. Message rules were unaffected (they were and remain correct); only the Bethe free energy contribution of a `softdot` node under full mean-field factorization was wrong, and only when `mean(q_γ) ≠ 1`. The mean-field variant now agrees with the (already-correct) structured `q(y, x)` variant. This bug was present unchanged from at least v4.4.3 through v6.3.1, so any Bethe free energy values previously reported for mean-field `softdot` models were biased by `½·mean(q_γ)·(mean(q_γ)−1)·(−2·m_y·m_θᵀm_x)` ([#615](https://github.com/ReactiveBayes/ReactiveMP.jl/issues/615))
+
 ## [6.3.1] - 2026-07-06
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Mean-field `@average_energy` for the `softdot` node no longer applies a spurious extra `mean(q_γ)` factor to the cross term `−2·m_y·m_θᵀm_x`, which biased the term by `mean(q_γ)²` instead of `mean(q_γ)`. Message rules were unaffected (they were and remain correct); only the Bethe free energy contribution of a `softdot` node under full mean-field factorization was wrong, and only when `mean(q_γ) ≠ 1`. The mean-field variant now agrees with the (already-correct) structured `q(y, x)` variant. This bug was present unchanged from at least v4.4.3 through v6.3.1, so any Bethe free energy values previously reported for mean-field `softdot` models were biased by `½·mean(q_γ)·(mean(q_γ)−1)·(−2·m_y·m_θᵀm_x)` ([#615](https://github.com/ReactiveBayes/ReactiveMP.jl/issues/615))
 
+### Changed
+- CI: the `test` job now runs with a 60-minute timeout so a hung dependency fails fast instead of consuming GitHub's default 6-hour limit ([#618](https://github.com/ReactiveBayes/ReactiveMP.jl/pull/618))
+
 ## [6.3.1] - 2026-07-06
 
 ### Fixed
--- a/src/annotations/input_arguments.jl
+++ b/src/annotations/input_arguments.jl
@@ -126,9 +126,13 @@ function post_product_annotations!(
             _merge_input_arguments(left_record, right_record),
         )
     elseif has_left
-        annotate!(merged, :rule_input_arguments, get_rule_input_arguments(left_ann))
+        annotate!(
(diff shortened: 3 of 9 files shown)


Nimrais:
I checked the PR, @mhidalgoaraya very good catch thanks. It helped me with one diverging free energy plot in my dissertation 🙇.

@bvdmitri If you want I can review the PR if Marco does not have time.

bvdmitri:
@Nimrais would be nice if you can approve indeed!

Nimrais (approved):
Good, I do not see problems.

The only thing that this PR is actually showing that maybe we need to add some more tests with MC comparison. We can sample from the resulting local marginal take average the energy and check that it's actually converging to the @average_energy result.

bvdmitri:
CI is stalling for some reason, I will investigate

bvdmitri:
CI on this PR is not failing because of these changes. The test suite hangs (all Julia versions hit the 6h CI timeout) due to an upstream regression in Roots.jl v3.0.2.

Root cause: Roots v3.0.2 regressed the `ITP()` bracketing solver. `Distributions.quantile` uses it, and the `BinomialPolya` (PolyaGamma) rule samples from a truncated `InverseGaussian`, which now receives `NaN` from `quantile` and spins forever in the sampler's accept/reject loop. `main` last passed on Jul 6 with Roots 3.0.1; 3.0.2 was released Jul 8. Nothing in this PR is involved (the hang reproduces on `main` too).

Upstream:
- JuliaMath/Roots.jl#515
- JuliaStats/Distributions.jl#2080

Blocked until the Roots fix is released, or until we add a temporary `Roots < 3.0.2` bound. Not merging yet.

PR #620, opened 2026-07-13
Fix structured VMP deadlock with 3+ mutually-dependent clusters  (#344)

### LINKED ISSUE #344: Initializing message towards node with SVMP with multiple clusters breaks inference procedure
This only works when we have 4 or more interfaces, since we need SVMP with 3 or more clusters. For example, in the GCV node, we have $q(x, y)q(w)q(k)q(z)$ as factorization constraint. Let's define the following model:
```julia
@model function demo(ay)
    x ~ NormalMeanVariance(0.0, 1.0)
    z ~ NormalMeanVariance(0.0, 1.0)
    k ~ NormalMeanVariance(0.0, 1.0)
    w ~ NormalMeanVariance(0.0, 1.0)
    y ~ GCV(x, z, k, w) 
    ay ~ NormalMeanVariance(y, 1.0)
end
c = @constraints begin
    q(x,y,z,k,w) = q(x, y)q(k)q(z)q(w)
end
```
Now in order to run inference in this model, we need to initialize the marginals on `k`, `z` and `w`, otherwise we cannot compute the messages towards `x` and `y`. This checks out:
```
i = @initialization begin
    q(k) = NormalMeanVariance(0.0, 1.0)
    q(w) = NormalMeanVariance(0.0, 1.0)
    q(z) = NormalMeanVariance(0.0, 1.0)
end
```
However, if we now initialize a message on `y`, the inference procedure breaks:
```julia
i = @initialization begin
    q(k) = NormalMeanVariance(0.0, 1.0)
    q(w) = NormalMeanVariance(0.0, 1.0)
    q(z) = NormalMeanVariance(0.0, 1.0)
    μ(y) = NormalMeanVariance(0.0, 1.0)
end
```
```
Variables [ w, k, z, x ] have not been updated after an update event. 
Therefore, make sure to initialize all required marginals and messages. See `initialization` keyword argument for the inference function. 
See the official documentation for detailed information regarding the initialization.
```

gauraaansh:
Fixes ReactiveBayes/RxInfer.jl#344.

```__collect_latest_updates``` combines sibling cluster dependencies via ```combineLatestUpdates``` with ```PushNew()``` which requires every sibling to emit a new value before firing again. If a sibling only ever emits a provisional ```(is_initial)``` value before settling on its real one, the mutual refresh requirement is never satisfied again once that stream settles, permanently deadlocking 3+ mutually dependent structured VMP clusters.

Fix mirrors the existing reset_vstatus pattern from clusters.jl/random.jl: when the combined sources are still provisional, force the wrapper's vstatus to refresh.

Verified: fixes the reported repro, converges to the same posterior as the unbroken case, no hang under a full-seed stress case, full existing test suite passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
-- Structured VMP with 3+ mutually-dependent clusters no longer deadlocks when a provisional (`is_initial`) sibling value permanently retired `PushNew()`'s mutual-refresh requirement once it settled on its real value ([#344](https://github.com/ReactiveBayes/RxInfer.jl/issues/344), [#620](https://github.com/ReactiveBayes/ReactiveMP.jl/pull/620))
+- Structured VMP with 3+ mutually-dependent clusters no longer deadlocks when an initialized message caused the first outbound message computation to consume only provisional (`is_initial`) marginals: the marginal dependencies of a node now stay consumable while all of their recent values are `is_initial`, instead of `PushNew()` requiring every dependency to refresh before the next computation. The relaxation applies to marginal dependencies only — message dependencies keep strict `PushNew()` semantics, so message-update schedules (and hence free-energy trajectories) of unaffected models remain unchanged ([#344](https://github.com/ReactiveBayes/RxInfer.jl/issues/344), [#620](https://github.com/ReactiveBayes/ReactiveMP.jl/pull/620))
 
 ## [6.3.2] - 2026-07-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Structured VMP with 3+ mutually-dependent clusters no longer deadlocks when a provisional (`is_initial`) sibling value permanently retired `PushNew()`'s mutual-refresh requirement once it settled on its real value ([#344](https://github.com/ReactiveBayes/RxInfer.jl/issues/344), [#620](https://github.com/ReactiveBayes/ReactiveMP.jl/pull/620))
+
 ## [6.3.1] - 2026-07-06
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


gauraaansh:
I kept this PR to just the fix. Please let me know if I should add the regression tests as well.

bvdmitri:
Thanks for working on this!! Yes, would definitely be nice to add a regression test. You can use the example from the issue itself and see that it works now :)

bvdmitri:
Thanks for the fix @gauraaansh — the diagnosis of the deadlock and the `reset_vstatus` direction are spot on! I looked into why the downstream `RxInfer.jl` CI job was failing and pushed two commits on top of your branch.

**Why the downstream job failed.** The failing test (`test/models/aliases/aliases_gamma_tests.jl`, `all(<=(1e-14), diff(results.free_energy))`) is fully deterministic, and the failure was caused by the reset being applied to *message* dependencies in addition to marginal dependencies. With the reset on message dependencies, an outbound message is recomputed as soon as a single dependency refreshes while the others are still `is_initial`. In the `gamma_aliases` model this fires at the binary `+` nodes (whose inbound messages all start as initial), which changes the free-energy trajectory from iteration 1. The model still converges to the identical posterior, but the FE approaches the fixed point from below and creeps back up by ~1e-13 near convergence, tripping the strict monotonicity assertion.

**What I changed.**
- Scoped `__reset_vstatus_of_sources` to marginal dependencies only. This is sufficient to fix the #344 deadlock (verified against the GCV repro from the issue), and it leaves the message-update schedule of unaffected models untouched — with this scoping the FE trajectory of the failing RxInfer test is bit-identical to `main`.
- Added two unit regression tests in `test/nodes/dependencies_tests.jl`: marginal-dependency combinations must re-fire while all recent values are `is_initial` (fails without your fix), and message-dependency combinations must keep strict `PushNew()` semantics (fails if the reset is applied to messages).
- Rewrote the CHANGELOG entry to describe the scoped behavior.
- Drive-by: fixed selective test runs (`make test test_args=...`) in `test/runtests.jl` — `map` had been missing its `ARGS` argument since the TestItemRunner migration.

**Verification.** Full ReactiveMP suite passes locally (1.12.6), full RxInfer suite passes against this branch including the previously failing `aliases for Gamma family of distributions` test, and the GCV repro from #344 runs to convergence with both initialization variants agreeing on the posteriors. A downstream regression test for the #344 scenario is prepared for RxInfer and will be PR'd there once this lands in a release.

bvdmitri:
The downstream regression test for the #344 scenario is now open as a draft PR on the RxInfer side: ReactiveBayes/RxInfer.jl#678. It passes against this branch and will be undrafted once a ReactiveMP release containing this fix is registered.

bvdmitri (approved):

Issue #625, opened 2026-08-02
MEDIUM: EqualityChain shares an unsynchronized mutable cache — race/lost-update hazard under AsyncScheduler


docxology:
## Summary

`EqualityChain` (built for every variable of degree>1, `variables/random.jl`) shares two `BitVector`s (`cacheleft`/`cacheright`) and per-node mutable `Message` caches among all outbound mappings. `materialize!` (`src/nodes/equality.jl:130-152`) and `ChainInvalidationCallback` (`equality.jl:169-182`) read/write those structures with **no lock**. Under the documented `ScheduleOnStreamPostprocessor(AsyncScheduler())` (`postprocessors/scheduled.jl`), concurrent outbound computations for different indices can tear/overwrite the cache and produce outbound messages from stale or partially-written prefix products — a silent-wrong-message hazard.

## Impact / scope

- Default single-threaded scheduler is safe; this is latent.
- Opt-in async scheduling + high-degree variables ⇒ potential lost updates → silent wrong messages.

## Suggested fix

Hold a per-chain `ReentrantLock` across `materialize!` and the invalidation callback (or use atomic cache bits and re-validate `iscached` before trusting a read). This is a design/trade-off change, so a PR description (not a committed branch) is drafted in `prs/pr-17-equalitychain-lock.md` for maintainer discussion.

bvdmitri:
Maintainer triage: **the premise of this report is incorrect — the described data race cannot occur — and the suggested fix would cost real performance to defend against it.** Leaving open as a note rather than closing outright, since the underlying "is the cache reasoning airtight?" question is a fair one to keep on record.

### `AsyncScheduler` is not multithreaded

The report's impact rests on "under threaded scheduling ... concurrent outbound computations for different indices can tear/overwrite the cache" and calls it "undefined behavior". But `AsyncScheduler` is a Rocket.jl scheduler built on coroutines, not threads. In `Rocket/src/schedulers/async.jl` every concurrency primitive is `@async`:

```
src/schedulers/async.jl:57:        @async begin
src/schedulers/async.jl:105:    channeling_task = @async begin
src/schedulers/async.jl:114:    subscription_task = @async begin
```

[...]

PR #639, opened 2026-08-02
Fix: audit inputargs copy

docxology:
**[Verified audit fix]** — branch `fix/audit-inputargs-copy`.

Commit: `Make InputArgumentsAnnotations merges copy-on-write`

## Changes

src/annotations/input_arguments.jl | 13 +++++++------
 src/message.jl                     |  5 ++++-
 src/rules/gcv/w.jl                 |  6 +++---
 3 files changed, 14 insertions(+), 10 deletions(-)

## Verification

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- `InputArgumentsAnnotations` no longer records rule executions that never contributed to a message. `_merge_input_arguments` merged by `push!`/`append!`/`pushfirst!`-ing into one operand's `mappings` vector and returning that same object, but the operand is owned by a `Message` that `EqualityChain` caches and reuses across several all-but-one outbound products. The second product therefore saw a record that had already grown from the first, so the stored trace accumulated entries from unrelated products. Merging is now copy-on-write: it builds a fresh `ProductInputArgumentsRecord` and leaves both operands untouched. This affects only the diagnostic trace returned by `get_rule_input_arguments`, never the messages themselves, and only when `InputArgumentsAnnotations` is enabled ([#622](https://github.com/ReactiveBayes/ReactiveMP.jl/issues/622))
+
 ## [6.3.3] - 2026-07-14
 
 ### Fixed
--- a/src/annotations/input_arguments.jl
+++ b/src/annotations/input_arguments.jl
@@ -88,22 +88,23 @@ end
 function _merge_input_arguments(
     left::RuleInputArgumentsRecord, right::ProductInputArgumentsRecord
 )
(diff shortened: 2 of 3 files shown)


bvdmitri:
Maintainer review — **the aliasing defect is real and the copy-on-write fix is correct**, though the issue's framing of it as a threading problem is not. I have rebased this branch and pushed a regression test plus a CHANGELOG entry (thanks for enabling maintainer edits).

### Rebased to a single concern

As submitted, this branch carried **three** commits — the GCV `:ω` fix, the `LogScaleAnnotations` fix, and this one — because #641 ⊂ #640 ⊂ #639 formed a cumulative stack. Merging it would have silently landed two unrelated changes. Rebased onto `main`, so it now contains only `src/annotations/input_arguments.jl` plus its test.

### The aliasing is real, and single-threaded

Confirmed: `_merge_input_arguments` merged by mutating one operand's `mappings` vector and returning that same object. That operand is owned by a `Message`, and `EqualityChain` deliberately caches messages and hands the *same* object to several all-but-one outbound products (`getcache`/`materialize!` in `src/nodes/equality.jl`). So the second product observed a record that had already grown during the first, and the stored trace accumulated rule executions that never contributed to it.

The new test reproduces exactly that: under the in-place implementation, the second product's trace came back as `[:a, :b, :c, :d]` — containing `:c`, from a product it never participated in.

### Correcting the threading claim

The issue lists as its second impact: *"Threaded (`AsyncScheduler`): unsynchronized mutation of shared `mappings` vectors is undefined behavior."* That does not apply. `AsyncScheduler` is a Rocket.jl scheduler built on `@async` — cooperatively-scheduled coroutines on a single thread. There is no `Threads.@spawn` anywhere in `Rocket/src/schedulers/async.jl`; it synchronizes through a `Channel`. Two such tasks never run simultaneously, so no shared-memory tearing can occur, and `_merge_input_arguments` contains no yield points either.

This matters for how the fix is justified: **the aliasing is the entire defect**, it needs no concurrency to trigger, and the impact is confined to the diagnostic trace returned by `get_rule_input_arguments` — messages themselves were never affected. The CHANGELOG entry I pushed describes it that way rather than repeating the data-race framing. (Same correction applies to #625, where the threading premise is the *only* claim; commented there.)

### Regression test added

Four testsets, one per formerly-mutating method plus one through the public `post_product_annotations!` entry point with a shared `left_ann`:

| `src/annotations/input_arguments.jl` | result |
|---|---|
| reverted to pre-fix `main` | **14 failed, 3 errored** |
| with this PR's fix | **18 pass** |

The pre-existing merge tests pass in both runs, so the new tests add the aliasing property without over-constraining the merge semantics.

### On the fix itself

`vcat`-based reconstruction is the right call here and I kept it as submitted. Worth noting it also makes the merge order explicit at each call site, which the previous `pushfirst!`-into-`right` version obscured.

PR #640, opened 2026-08-02
Fix: audit logscale missing

docxology:
**[Verified audit fix]** — branch `fix/audit-logscale-missing`.

Commit: `Skip post-rule annotation processors when rule short-circuits to missing`

## Changes

src/message.jl     | 5 ++++-
 src/rules/gcv/w.jl | 6 +++---
 2 files changed, 7 insertions(+), 4 deletions(-)

## Verification

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- A `missing` message no longer crashes when `LogScaleAnnotations` is enabled. `MessageMapping` short-circuits to `missing` without running the rule when any input is `missing` (the mechanism behind `new_observation!(datavar, missing)`), but it still invoked the post-rule annotation processors afterwards. `LogScaleAnnotations.post_rule_annotations!` only defaults `:logscale` to `0` when *all* inputs are `PointMass`, and otherwise calls `error(...)` — so a deferred message from a node with non-`PointMass` inputs was turned into a hard failure, defeating the `missing` short-circuit entirely. Post-rule processors are now skipped when the rule was short-circuited, leaving the message genuinely deferred and annotation-free. Products involving such a message were already handled correctly by `post_product_annotations!`'s `::Missing` dispatches, which copy the other side's annotations through unchanged ([#623](https://github.com/ReactiveBayes/ReactiveMP.jl/issues/623))
+
 ## [6.3.3] - 2026-07-14
 
 ### Fixed
--- a/src/message.jl
+++ b/src/message.jl
@@ -689,7 +689,10 @@ function (mapping::MessageMapping)(messages, marginals)
         end
 
     # Run annotation processors after the rule has been executed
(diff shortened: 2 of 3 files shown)


bvdmitri:
Maintainer review — **the defect is real and the one-line fix is correct.** I have rebased this branch and pushed a regression test plus a CHANGELOG entry (thanks for enabling maintainer edits).

### Rebased to a single concern

As submitted, this branch sat on top of #641's GCV commit, so its diff included `src/rules/gcv/w.jl` — an unrelated change. (#641 ⊂ #640 ⊂ #639 was the full stack; merging #639 would have landed all three at once.) Rebased onto `main`, so it now contains only the annotation change plus its test.

### Verified reachable

Confirmed by code trace: `MessageMapping` sets `result = missing` and skips the rule when any inbound message or marginal is `missing` (`src/message.jl:657-663`) — the mechanism behind `new_observation!(datavar, missing)` — then runs `post_rule_annotations!` regardless. `LogScaleAnnotations.post_rule_annotations!` only defaults `:logscale = 0` when *all* inputs are `PointMass` (`logscale.jl:47-49`); a `missing` message is not, so it falls to `error(...)`. A legitimately deferred message became a hard crash, defeating the short-circuit.

### One correction to the reasoning, in the fix's favour

I initially expected this fix to be **insufficient** — a `missing`-data message still reaches `compute_product_of_two_messages`, which calls `post_product_annotations!` unconditionally (`message.jl:236`), and `LogScaleAnnotations.post_product_annotations!` reads `getlogscale` on *both* sides (`logscale.jl:68-69`). That looked like it would simply trade the `error` for a `KeyError(:logscale)` one layer down.

It does not, and the reason is worth recording: the generic `post_product_annotations!` already dispatches on `::Missing` for `left_dist`/`right_dist` (`src/annotations.jl:168-190`) and copies the other side's annotations through **without ever consulting the per-processor method**. So `LogScaleAnnotations` is never reached with an absent `:logscale`. The fix is sufficient as written — but that was load-bearing behaviour nobody had pinned, so the test now covers it explicitly in both orderings (missing-on-left and missing-on-right).

### Regression test added

`test/annotations/logscale_tests.jl`, three testsets deliberately arranged so the important one cannot pass vacuously:

1. **Control**: a *concrete* non-`PointMass` message still reaches the processor and still errors. Without this, the `missing` test would also pass if the processor had simply stopped running altogether.
2. **The regression**: a `missing` message comes back as `missing` and carries no `:logscale` — no log-scale is invented for a message that never went through a rule.
3. **Downstream**: the deferred message multiplied against a concrete one, both orderings, asserting the concrete side's log-scale propagates.

| `src/message.jl` | result |
|---|---|
| reverted to pre-fix `main` | **2 errors** (testsets 2 and 3) |
| with this PR's fix | **5 pass** |

The control testset passes in both runs.

### Note on the submitted description

The PR body listed `src/rules/gcv/w.jl` under "Changes" and said "Confirmed against `main` in the ReactiveBayes audit ... sources verified read-only", but carried no test. For a crash on a documented code path in the annotations layer, a test is the deliverable — now added.

Issue #626, opened 2026-08-02
MAJOR: GCV message rules are completely untested — silent ω-corruption fix ships with no regression test


docxology:
## Summary

The whole **GCV** message-rule family (`src/rules/gcv/{w,k,y,x,z,marginals,gaussian_extension}.jl`) has **zero direct test coverage**. The only GCV test (`test/nodes/predefined/gcv_tests.jl`) validates the `ExponentialLinearQuadratic` distribution statistics and `prod` — it never calls a single message rule. There is no `test/rules/gcv/` directory.

This is not academic: the tip commit `c0aafcea` ("Fix GCV(:ω) mean-field rule copy-paste of :κ coefficients") documents that the mean-field `GCV(:ω)` rule *"silently corrupts the ω posterior whenever GCV runs with a mean-field variational family."* That is exactly the class of bug a message-rule test catches — and the fix lands with **no accompanying test anywhere**, so the same defect class can recur undetected.

**Package/version:** ReactiveMP.jl 6.3.3

## Evidence

```
$ grep -rln "GCV" test/                      # zero hits — no test file even names the node
$ grep -rn "@rule GCV\|@marginalrule GCV\|@average_energy GCV" test/   # zero hits
$ ls test/rules/ | grep gcv                  # no test/rules/gcv/ directory
```

The only GCV test (`test/nodes/predefined/gcv_tests.jl`) validates `ExponentialLinearQuadratic`
statistics and `prod` — it never calls a single message/marginal rule.

Related: four more `src/rules/` families define message rules but have no `@test_rules` coverage:
- `gamma` — `src/rules/gamma/{out,marginals}.jl`
- `mixture` — `src/rules/mixture/{inputs,out,switch}.jl` (node test is purely structural)
- `uninformative` — `src/rules/uninformative/out.jl` (only `prod` tested)
- `dirichlet_collection` — `src/rules/dirichlet_collection/{out,marginals}.jl` (only used as an input value elsewhere)

## Suggested fix

1. Add `@test_rules` entries for every GCV rule signature (factorized and structured variants of `:ω`, `:κ`, `:y`, `:x`, `:z`) with hand-computed expected outputs.
2. Add `@test_marginalrules` for `GCV(:y_x)`.
3. **Regression:** pin the fixed mean-field `:ω` coefficients `(a,b,c,d) = (1, ψ·E[e^{-κz}], -1, 0)` (matching the structured rule) so the copy-paste cannot silently return.
[...]

PR #641, opened 2026-08-02
Fix: gcv omega meanfield copypaste

docxology:
**[Verified audit fix]** — branch `fix/gcv-omega-meanfield-copypaste`.

Commit: `Fix GCV(:ω) mean-field rule copy-paste of :κ coefficients`

## Changes

src/rules/gcv/w.jl | 6 +++---
 1 file changed, 3 insertions(+), 3 deletions(-)

## Verification

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- The mean-field `GCV(:ω)` message rule (the `q_y`/`q_x` factorized method) was a copy-paste of the `:κ` rule and stamped `z`-dependent coefficients `(a, c, d) = (⟨z⟩, −⟨z⟩, Var(z))` onto the `ω` message. Since the `GCV` node is `p(y | x, z, κ, ω) = N(y | x, exp(κz + ω))`, `ω` enters the log-likelihood linearly with a *constant* coefficient of one and never inside the exponential's quadratic slot, so the correct coefficients are `(1, ψ·⟨e^{−κz}⟩, −1, 0)` — none of which depend on `z`. The bug silently corrupted the `ω` posterior, and everything downstream of it, whenever `GCV` ran under a mean-field factorization; it produced no error and no warning. The structured `q(y, x)` method and both `:κ` methods were and remain correct, so the two `:ω` methods disagreed with each other — an invariant now pinned by a regression test. `GCV` message rules previously had no test coverage at all, which is how this shipped undetected ([#621](https://github.com/ReactiveBayes/ReactiveMP.jl/issues/621))
+
 ## [6.3.3] - 2026-07-14
 
 ### Fixed
--- a/src/rules/gcv/w.jl
+++ b/src/rules/gcv/w.jl
@@ -33,10 +33,10 @@ end
     A = exp(-κ_mean * z_mean + γ / 2)
     psi = (y_mean - x_mean)^2 + y_var + x_var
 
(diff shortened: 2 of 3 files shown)


bvdmitri:
Maintainer review — **the fix is correct, and I verified it independently rather than taking the PR description's word for it.** I have pushed a regression test and a CHANGELOG entry to this branch (thanks for enabling maintainer edits).

### Independent verification of the coefficients

Deriving from the node's own `@average_energy` (`src/nodes/predefined/gcv.jl:95-137`), which pins the likelihood as `p(y | x, z, κ, ω) = N(y | x, exp(κz + ω))` — i.e. the **variance** is `exp(κz + ω)`:

```
−log p = ½[log2π + (κz + ω) + (y − x)²·e^{−(κz+ω)}]
```

Marginalising `y, x, z, κ` and keeping the `ω`-dependent terms:

```
−log f(ω) = ½[ω + ψ·⟨e^{−κz}⟩·e^{−ω}] + const,    ψ = ⟨(y − x)²⟩
```

`ExponentialLinearQuadratic(a,b,c,d)` has `logpdf = −(a·ω + b·exp(c·ω + d·ω²/2))/2`, so matching term by term gives `a = 1`, `b = ψ·⟨e^{−κz}⟩`, `c = −1`, `d = 0`. Exactly what this patch sets. ✅

The structural reason for the bug is worth recording: `:ω` is the *only* one of the three `ExponentialLinearQuadratic`-returning rules where the variable's linear coefficient is the constant `1` rather than another variable's mean. `:κ` correctly carries `(⟨z⟩, −⟨z⟩, Var(z))` and `:z` correctly carries `(⟨κ⟩, −⟨κ⟩, Var(κ))` — both are right, and both are symmetric in a way `:ω` is not. Copy-pasting from `:κ` therefore produced something that *looks* structurally plausible next to its siblings, which is why review missed it.

### Regression test added

`test/rules/gcv/w_tests.jl` (new directory — there was no `test/rules/gcv/` at all). Four independent checks:

1. Both methods against an in-source reference derived from the average energy, not transcribed from the implementation. `z` deliberately gets `⟨z⟩ ≠ 1` and `Var(z) ≠ 0`, since with `⟨z⟩ = 1, Var(z) = 0` the bug is invisible.
2. Explicit `a == 1`, `c == -1`, `d == 0`.
3. **The cross-method invariant**: with block-diagonal `q(y, x)` the structured variant reduces exactly to the mean-field one, so the two must return identical coefficients. This is the check that most directly encodes the defect.
4. A direct anti-regression that `:ω` does not share `(a, c, d)` with `:κ`.

**Verified the test actually catches the bug** rather than just passing:

| `src/rules/gcv/w.jl` | result |
|---|---|
| reverted to pre-fix `main` | **18 failures** |
| with this PR's fix | **32 pass** |

The structured-variant testset passes in *both* runs, which confirms the test isolates the mean-field method instead of over-constraining the node.

### Notes on the PR as submitted

- The description said the fix was "confirmed ... sources verified read-only", but shipped no test — for a silent numerical corruption in an entirely untested rule family, that is the one thing that was actually needed. Now added.
- No CHANGELOG entry; added, spelling out the impact and that the structured method was unaffected.
- This branch is clean and single-concern. Note that #640 and #639 are **cumulative stacks on top of this commit** (#641 ⊂ #640 ⊂ #639), so merging #639 as-is would silently land all three unrelated changes. I am rebasing those two to their own single concerns separately.

Remaining coverage for the rest of the `GCV` rule family (`:y`, `:x`, `:z`, `:κ`, the `:y_x` marginal, and both average-energy methods) is tracked in #626 and will follow in its own PR.

Issue #621, opened 2026-08-02
MAJOR: GCV(:ω) mean-field rule is a copy-paste of the :κ rule → wrong ω posterior in mean-field mode


docxology:
## Summary

The mean-field (`q_y`, `q_x` factorized) variational message to `ω` of the **GCV** (Gaussian chain variational) node is a copy-paste of the `:κ` rule. It stamps `z`-dependent coefficients on the `ω` message, silently corrupting the posterior of `ω` (and everything downstream) whenever GCV runs with a factorized variational family. The structured variant and both `:κ` variants are correct — only this one method is broken.

**Package/version:** ReactiveMP.jl 6.3.3 · Julia 1.12.6

## Location

`src/rules/gcv/w.jl:24-42`, specifically lines 36-39:

```julia
a = z_mean
b = psi * A
c = -a          # = -z_mean
d = z_var
```

The correct message-to-`ω` (marginalising `y,x,z,κ`) has coefficients `(a,b,c,d) = (1, ψ·E[e^{-κz}], -1, 0)`. The structured sibling `w.jl:16-19` already implements this correctly (`a=1, c=-1, d=0`). The factorized method instead uses the `:κ` coefficients (compare `src/rules/gcv/k.jl:34-38`).

## Repro

```julia
using ReactiveMP, Distributions, BayesBase, ExponentialFamily
q_y = NormalMeanVariance(3.0, 1.0); q_x = NormalMeanVariance(1.0, 2.0)
q_yx = MvNormalMeanCovariance([3.0,1.0], [1.0 0.3; 0.3 2.0])
q_z = NormalMeanVariance(0.5, 0.7); q_k = NormalMeanVariance(0.8, 0.4)
q_w = NormalMeanVariance(1.2, 0.5)
meta = ReactiveMP.GCVMetadata(ReactiveMP.GaussHermiteCubature(20))

wf = @call_rule GCV(:ω, Marginalisation) (q_y=q_y, q_x=q_x, q_z=q_z, q_κ=q_k, meta=meta)
ws = @call_rule GCV(:ω, Marginalisation) (q_y_x=q_yx, q_z=q_z, q_κ=q_k, meta=meta)
```

Observed:

```
wf (a,b,c,d) = (0.5, 7.099, -0.5, 0.7)   # WRONG (matches :κ, z-dependent)
ws (a,b,c,d) = (1.0, 6.490, -1.0, 0.0)   # CORRECT
```

[...]

PR #645, opened 2026-08-10
Add test coverage for the GCV message rule family

bvdmitri:
Addresses #626.

## Problem

The `GCV` message rules had **no direct coverage whatsoever**:

```
$ grep -rn "GCV" test/          # nothing
$ ls test/rules/ | grep gcv     # no such directory
```

The one GCV test file (`test/nodes/predefined/gcv_tests.jl`) exercised the `ExponentialLinearQuadratic` distribution and its `prod`, but never called a message rule. That is precisely how the mean-field `:ω` copy-paste (#621) shipped undetected.

## Change

New `test/rules/gcv/` covering `:y`, `:x`, `:z`, `:κ` in both mean-field and structured forms, the `:y_x` joint marginal, and both `@average_energy` methods.

Every expected value is **derived in-source** from the node's likelihood `N(y | x, exp(κz + ω))` — which the node's own `@average_energy` independently pins — rather than captured as a golden number. The derivations are written out in comments so a reader can audit them without running the code.

### The checks that carry the most weight need no reference values at all

They assert that two independent code paths agree, so they cannot be defeated by a reference that was mis-derived in the same direction as a bug:

- **Structured reduces to mean-field at `Cov(y, x) = 0`** — the invariant #621 violated.
- **`:y` ≡ `:x` under `y ↔ x`**, and **`:κ` ≡ `:z` under `κ ↔ z`** — the likelihood sees only `y - x` and the product `κz`.
- **The `:y_x` marginal's off-diagonal coupling is the reciprocal of the variance the `:y` rule adds** — two code paths computing the same effective noise precision.
- **The two `@average_energy` methods agree at zero covariance** — the same invariant that caught the softdot bug in #615.
- **`:κ` and `:z` must not coincide**, and each must carry the moments of the variable it actually marginalised.

Plus structural properties: the joint precision stays positive-definite, the marginal means bracket the two incoming means, and both `ExponentialLinearQuadratic` rules satisfy `a = -c` by construction.

### Two node behaviours now documented by tests rather than left implicit

`shared_tests.jl` establishes what the recurring expectations actually are:

- `A = ⟨e^{-ω}⟩` is **exact** — the mean of a lognormal — confirmed against an independent Monte Carlo estimate.
- `B = ⟨e^{-κz}⟩` is a **deliberate Gaussian-moment approximation**. `κz` is a product of independent Gaussians and is not itself Gaussian, so the lognormal formula does not give its exact exponential moment. The test asserts the approximation is in the right ballpark **and explicitly that it is not exact**, so a future change to an exact computation surfaces as a decision to revisit rather than passing silently.

## Scope note

The `:ω` rule's coverage is **not** here — it lives in `test/rules/gcv/w_tests.jl` on #641, together with the fix it depends on. This PR is based on `main` and is green on `main` as-is, so the two can merge in either order. Once both land, `w_tests.jl` and `shared_tests.jl` are complementary and cross-reference each other in comments.

## Result

GCV assertions go from **57** (distribution only) to **416**, all passing:

```
test/rules/gcv/y_tests.jl          |  85
test/rules/gcv/x_tests.jl          |  73
test/rules/gcv/k_tests.jl          |  19
test/rules/gcv/z_tests.jl          |  19
test/rules/gcv/shared_tests.jl     |  19
test/rules/gcv/marginals_tests.jl  | 133
test/nodes/predefined/gcv_tests.jl |  68
```

Tests only — no source changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Test coverage for the `GCV` message rule family, which previously had none — `grep -rn GCV test/` matched nothing, and the only GCV test file exercised the `ExponentialLinearQuadratic` distribution and `prod` without ever calling a message rule. New `test/rules/gcv/` covers `:y`, `:x`, `:z`, `:κ` (mean-field and structured variants), the `:y_x` joint marginal, and both `@average_energy` methods, with expected values derived in-source from the node's likelihood `N(y | x, exp(κz + ω))` rather than captured as golden numbers. Includes cross-method invariants that need no reference values at all: structured variants must reduce to their mean-field siblings at zero `y`–`x` covariance, `:y` and `:x` must agree under the node's `y ↔ x` symmetry, `:κ` and `:z` must agree under `κ ↔ z`, and the joint marginal's off-diagonal coupling must be the reciprocal of the variance the `:y` rule adds. GCV assertions go from 57 (distribution only) to 416 ([#626](https://github.com/ReactiveBayes/ReactiveMP.jl/issues/626))
+
 ## [6.3.3] - 2026-07-14
 
 ### Fixed
--- a/test/nodes/predefined/gcv_tests.jl
+++ b/test/nodes/predefined/gcv_tests.jl
@@ -83,3 +83,146 @@
         end
     end
 end
+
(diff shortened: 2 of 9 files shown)


bvdmitri:
Superseded by #658, merged into main, which contains this PR's GCV test suite verbatim (byte-identical files) plus the log-space source fix in the same GCV rules. Closing as subsumed.

PR #661, opened 2026-08-27
Adding GaussianCoupling node

Nimrais:
PR adds a new node in ReactiveMP: `Bilinear` node. The idea is based on this paper https://arxiv.org/html/0810.1119v1.

This node allows to do a cool stuff in RxInfer: solving the system of linear equations via message passing.

Example:
```julia
using RxInfer

A = [1.0 0.5; 0.5 4.0]
b = [5.0, 6.0]

@model function solve_linear_system(A, b)
    local x
    for i in 1:size(A, 1)
        x[i] ~ NormalMeanPrecision(b[i]/A[i, i], A[i, i])
    end
    for i in 1:size(A, 1)
        for j in (i+1):size(A, 2)
           x[j] ~ GaussianCoupling(x[i], -A[i, j])
        end
    end
end

result = infer(
    model = solve_linear_system(A=A),
    data = (b=b,),
    free_energy = true
)

@show mean.(result.posteriors[:x])
@show A \ b
# Wow you solved the linear system using RxInfer!
# But you also have uncertainty in your solution by the the variances.
@show var.(result.posteriors[:x])
```


bvdmitri:
Pushed a follow-up cleanup in 40da1f55: restored the pinned scripts manifest, expanded the GaBP documentation (supported factorization, convergence conditions, and the tree-only variance guarantee), tightened the average-energy signature, and migrated/expanded the rule tests with type-promotion, negative-coupling, properness-boundary, and end-to-end linear-system coverage. The focused Bilinear suite passes all 260 tests and the branch has been formatted with `make format`.

One remaining naming suggestion for Mykola: consider renaming `Bilinear` before release. Since this node represents an unnormalized pairwise Gaussian coupling rather than the usual bilinear map, `GaussianCoupling` or `BilinearPotential` may communicate the semantics more clearly and avoid collisions with Flux/Lux's `Bilinear`.

Nimrais:
I agree `GaussianCoupling` is better name I changed to it.

bvdmitri (approved):

PR #666, opened 2026-09-21
ManyPlus Node

Nimrais:
The idea is simple to have a node where you can iterate for some loop and write `y := ManyPlus([x[i] for i in ...])` without creating the itermediate variables which are very cubersome to init in RxInfer.

Semantically it is the same to write:
```Julia
...
for i in ...
      y := y + x[i]
end
```

But such construction as latter one is actually very hard to use in RxInfer.


bvdmitri:
@Nimrais, one question: could `ManyPlus` support an observed or fixed output, as in `y ~ ManyPlus(inputs = x)` with `y` given as data? Right now it fails with `RuleMethodError` because there's no rule for a `PointMass` message on `out`.

**Messages.** The rule would mirror the existing one:

```julia
@rule ManyPlus((:inputs, k), Marginalisation) (
    m_out::PointMass{<:Real},
    m_inputs::ManyOf{N, Union{UnivariateNormalDistributionsFamily, PointMass{<:Real}}},
) where {N} = begin
    μ = sum(m -> mean(m), m_inputs)
    v = sum(m -> var(m), m_inputs)
    return all(m -> m isa PointMass, m_inputs) ? PointMass(mean(m_out) - μ) : NormalMeanVariance(mean(m_out) - μ, v)
end
```

**Free energy.** When the output is fixed, the Gaussian inputs lie on the line where their sum equals `y` minus the constants, so their joint belief has `d − 1` dimensions. If you remove the last input using the sum constraint, the precision over the remaining `d − 1` inputs is `diag(1/v₁, …, 1/v_{d−1}) + 11ᵀ/v_d`. Its log-determinant simplifies to a symmetric form:

```
logdet Λ' = log(Σᵢ vᵢ) − Σᵢ log vᵢ
−H        = (logdet Λ' − (d − 1)(1 + log 2π)) / 2
```

For `d = 2` this gives `log((v₁ + v₂) / (v₁ v₂))`, which is the known binary result. The point-mass count would be one extra, for the observed output.

Note that binary `+` computes messages with an observed output, but its free energy fails (`MarginalRuleMethodError`). So the message rule alone would bring `ManyPlus` level with `+`, and the free-energy term would go beyond it.

Would you be up for adding this here, or would you rather do it as a follow-up?

---

Also, I pushed a docs commit to this branch (60844e01): it notes univariate-only support, documents how compile and run time scale with the number of inputs, adds the missing `local x` to the loop example (it didn't run without it), links GraphPPL instead of naming an RxInfer version, and moves the page to the end of the node list.

Nimrais:
> @Nimrais, one question: could ManyPlus support an observed or fixed output, as in y ~ ManyPlus(inputs = x) with y given as data? Right now it fails with RuleMethodError because there's no rule for a PointMass message on out

@bvdmitri Yeah I will add the pointmass rule, I see it's improssible. I just have not thought about it. 

> Also, I pushed a docs commit to this branch (https://github.com/ReactiveBayes/ReactiveMP.jl/commit/60844e01f935e40aaffed18edaa3975c91806876): it notes univariate-only support, documents how compile and run time scale with the number of inputs, adds the missing local x to the loop example (it didn't run without it), links GraphPPL instead of naming an RxInfer version, and moves the page to the end of the node list.

Ah yes thanks, I forgot it when I wrote it.

Nimrais:
@bvdmitri the review is addressed

PR #670, opened 2026-09-23
Generic mean-field test and how to use it

Nimrais:
I got a bit tired to see stuff like this https://github.com/ReactiveBayes/ReactiveMP.jl/issues/669. All our rules tested through some magic constants that it seems got selected just to make rules pass. I mean sometimes it's fine. But I would like to see a more techical approach to this problem.

We do not need at least for the full mean-field around stochastic nodes to relay on the LLMs to find it for us.

I have implemented a generic rule that is expected to be used only in tests, that is based on the fact that full mean-field rule is 
```math
\widehat{\log \mu}_{f \rightarrow i}(x_i) =
\frac{1}{N}\sum_{s=1}^{N}
\log f(x_1^{(s)}, \ldots, x_i, \ldots, x_K^{(s)}),
\qquad x_j^{(s)} \sim q_j \quad (j \ne i).
```

I added a test example that shows how to find the bug from the issue with it.

The impelementation is naïve if you want to improve it you are welcome.


bvdmitri:
FYI this class of errors will be solved in https://github.com/ReactiveBayes/ReactiveMP.jl/tree/refactor/rule-node-system-rewrite 

with a dedicated extensive testing library https://github.com/ReactiveBayes/ReactiveMP.jl/tree/refactor/rule-node-system-rewrite/lib/MessagePassingRulesTestUtils

bvdmitri:
particularly what you want also has been implemented [here](https://github.com/ReactiveBayes/ReactiveMP.jl/blob/ab8fb327185ff86b4ff2b0cde83fb77b367c6df2/lib/MessagePassingRulesTestUtils/src/verification.jl#L72) and this is how the bug has been discovered

